PR #2, opened 2015-07-19
Memory history goForward fixed

michalkvasnicak:
Push state was not saving state to entries so `goForward()` always failed on index check.

### DIFF
--- a/modules/__tests__/describeHistory.js
+++ b/modules/__tests__/describeHistory.js
@@ -1,6 +1,6 @@
 import assert from 'assert';
 import expect from 'expect';
-import { PUSH, REPLACE } from '../Actions';
+import { PUSH, REPLACE, POP } from '../Actions';
 
 function describeHistory(createHistory) {
   describe('when the user confirms a transition', function () {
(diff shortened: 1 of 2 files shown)


mjackson:
Thank you! Next big thing we need is a better testing stack so we can really flesh out the test suite and remove those describe.skip calls.

michalkvasnicak:
I have fixed browser history too (I guess). Should I create a PR so you can take a look and decide if it is OK? 

HashHistory is timing out, it seems that `ignoreNextHashChange` is the cause.

mjackson:
Sure! I can also make you a collaborator if you'd like so you don't have to wait on me to push fixes.

Though, as I said, until we can get something better than jsdom I won't feel comfortable making a ton of changes. If we can start testing in real browsers we can probably move much more quickly. :)

michalkvasnicak:
Thank you but I don't know, I don't want to make some mistakes and cause trouble :D 

I'll take a look on testing in browsers, but I don't have much experience with that. :)

PR #3, opened 2015-07-19
Browser history goBack/goForward fix

michalkvasnicak:
Location state was not saved so `getCurrentLocation()` always returned `null` for given `location.key`.

Really don't know if I get it right. :) Should only a key be saved to `window.history.state` as it is, or should whole location be saved instead? Now it is saving key to `window.history.state` and location is saved using `saveState()`.


mjackson:
@michalkvasnicak Can we get a simple test here, please?

michalkvasnicak:
@mjackson test from test suite was not passing. Now if you allow skipped test for goBack, goForward, it will pass. Is it enough? :)

michalkvasnicak:
This is passing for #6 too (locally in Chrome). HashHistory is still failing.

mjackson:
Superseded by #6, closing.

Issue #5, opened 2015-07-20
Better testing stack


mjackson:
It's pretty clear to me at this point that [jsdom](https://github.com/tmpvar/jsdom) isn't going to work. The support for the HTML5 history API appears to be incomplete, and, given the quirks of the various history implementations, I'd rather test in real browsers anyway.

Stuff we need to be able to do in a single integration test:

- Get reliable `hashchange`/`popstate` events when the URL changes
- Use `window.history.pushState`/`window.history.replaceState`
- Use `window.location.replace`
- Assign the value of `window.location.href`

Stuff we need to be able to do in a test suite:

- Run multiple integration tests without leaking state between them (see https://github.com/rackt/history/issues/5#issuecomment-122992242)

There are a few options out there, none of which I'm very familiar with. If anyone has any experience with any of these stacks, please chime in.

- [karma](http://karma-runner.github.io/0.13/index.html)
- [webdriver.io](http://webdriver.io/) (Selenium 2.0 bindings for node)
- ~~[intern](https://theintern.github.io/) (Doesn't seem very flexible, but could maybe work)~~
- [zuul](https://github.com/defunctzombie/zuul) (Powered by [electron](https://github.com/atom/electron))

We use karma to test [React Router](https://github.com/rackt/react-router), so it would be great if we could stick with that. But it seems like I remember it complaining when we tried to change the URL (which is strange because I thought karma ran all your tests in an `<iframe>` ... ?)

phated:
The library I was talking about on twitter - https://github.com/defunctzombie/zuul - it's my favorite runner, supporting mocha, jasmine, qunit frameworks, using the `--electron` flag will give you an electron instance to run your tests in and (I believe) gives you all of the stuff needed for your integration tests.  Changing the location doesn't seem to affect the reporter, so that's a plus.

phated:
I don't think you should use intern, I've been following it since release and it was promising but it is very inflexible and have heard nothing but complaints from anyone that tries to use it.

phated:
Also, you can use your webpack config as part of zuul using https://www.npmjs.com/package/zuul-builder-webpack

Dignifiedquire:
I'm biased (current maintainer of Karma) but I will suggest to use Karma. There is an issue with changing urls when using Karma, which I'm currently investigating and hope to have a fix for soon (Ref: https://github.com/karma-runner/karma/issues/1101)

mjackson:
[...]

PR #6, opened 2015-07-20
Zuul testing and history fixes

michalkvasnicak:
Added zuul as testing stack. Run `npm run test:zuul:local` and then open `http://localhost:8080/__zuul`.

* Fixed memory history state replacing
* Fixed HashHistory (**needs review, because I am stripping query key from location.search**)
* Added "complex" test for history `replaceState(), pushState(), goBack(), goForward()` using together


michalkvasnicak:
`window.location.href` is commented out (from beforeEach) because it is causing `Cannot GET /` in browser. Maybe someone can advise how to fix that? #5

michalkvasnicak:
Ok now is hash history fully passing tests. With #3 it seems that all tests are passing.

michalkvasnicak:
I added "complex" test for states and goBack/goForward because I found bug when integrating to [routex](https://github.com/michalkvasnicak/routex).

`replaceState` was acting like `pushState` in MemoryHistory. Now it is fixed but HashHistory and BrowserHistory are failing.

michalkvasnicak:
Ok now are all tests passing. You can close #3 in favour of this PR because I added same fix here.

mjackson:
Awesome! I was going to do this exact thing this morning! Now, instead of figuring all this out myself I can review your PR. :dancer: 

(also, I invited you as a collaborator @michalkvasnicak)

gaearon:
Happy to see you working together!

michalkvasnicak:
Thank you for invitation. :) I am glad I helped.

modules/createHashHistory.js
```diff
@@ -42,6 +46,7 @@ function createHashHistory(options={}) {
     var key, state;
     if (queryKey) {
       key = getQueryStringValueFromPath(path, queryKey);
+      path = stripQueryStringValueFromPath(path, queryKey);
```
mjackson: :+1: I was going to do this myself. Great change!

modules/createMemoryHistory.js
```diff
@@ -5,11 +5,11 @@ import createHistory from './createHistory';
 
 function createStorage(entries) {
   return entries
-    .filter(entry => entry.state)
-    .reduce((memo, entry) => {
-      memo[entry.key] = entry.state;
-      return memo;
-    }, {});
+      .filter(entry => entry.state)
```
mjackson: Let's use 2 space indent everywhere to keep things consistent :)

modules/createMemoryHistory.js
```diff
@@ -105,7 +105,11 @@ function createMemoryHistory(options={}) {
         current += 1;
         // fall through
```
mjackson: This comment is no longer relevant.

modules/createHashHistory.js
```diff
@@ -33,7 +37,7 @@ var DefaultQueryKey = '_k';
 function createHashHistory(options={}) {
   var { queryKey } = options;
 
-  if (queryKey && typeof queryKey !== 'string')
+  if (typeof queryKey !== 'string')
```
mjackson: I'd prefer to leave this line as it was so people can do `createHashHistory({ queryKey: true })` and get the default.

michalkvasnicak: This is bug, because if I don't pass anything then queryKey is undefined and DefaultQueryKey is no set.

michalkvasnicak: Or should it force people to specify queryKey in options? I though I can call just `createHashHistory()` and it's all set :)

mjackson: Ah, I see why you did this now. My original intent was to let people use hash history without a `queryKey`. They wouldn't be able to use state, but they would have somewhat cleaner URLs too. So it's a tradeoff.

michalkvasnicak: It can be done with the original line but then we need to add check to finishTransition to not save state if queryKey is undefined, because it will fail tests.

michalkvasnicak: But MemoryHistory and BrowserHistory are saving state by default, should HashHistory be saving it too? In that case maybe pass false as query key to disable saving state? `createHashHistory({ queryKey: false })` or something like that?

mjackson: Sure, `queryKey: false` to disable saving state sounds like a great idea.

mjackson:
I'm still not 100% sure that zuul is the best option for testing (see the discussion in #5) but it sure is better than what we have with jsdom so let's go ahead and merge. I'm going to keep working on getting a webdriver-based stack going, but this works great for now.

Thanks again for taking this on @michalkvasnicak! I'll give you the pleasure of pushing the merge button whenever you think you're ready :)

modules/createHashHistory.js
mjackson: Just FYI, I've put this in [the docs](https://github.com/rackt/history/blob/c5c71abf612d322388b11d352496e412a6900c16/docs/README.md#caveats-of-hash-history).

mjackson:
Sorry, just couldn't wait :)

michalkvasnicak:
It's ok, here is UTC+02:00 so I was sleeping :D

PR #11, opened 2015-07-22
Initial POP event has location.key defined

michalkvasnicak:
There is an ugly hack. 

Instead of calling `updateLocation` after registering listener and locations is undefined, we are replacing a state using the `replaceState` and in the `updateLocation` we are checking if a previous location is defined. If it isn't we replace action in a new location from REPLACE to POP and store it. Otherwise we are keeping it as it is.

We need it like this, because user expects initial event to be POP action and not REPLACE.

### DIFF
--- a/modules/__tests__/describeInitialLocation.js
+++ b/modules/__tests__/describeInitialLocation.js
@@ -1,0 +1,60 @@
+import expect from 'expect';
+import { POP } from '../Actions';
+import createMemoryHistory from '../createMemoryHistory';
+
+function describeInitialLocation(createHistory) {
+  describe('location has key on initial pop', function() {
+    var unlisten, history;
+
+    beforeEach(function() {
+      history = createHistory();
+    });
+
+    afterEach(function() {
+      if (unlisten) unlisten();
+    });
+
+    it('replaces state if location key is missing', function(done) {
+      unlisten = history.listen(function(location) {
+        try {
+          expect(location.action).toEqual(POP);
+          expect(location.key).toNotEqual(null);
+          expect(location.state).toEqual(null);
+          done();
+        } catch (e) {
+          done(e);
+        }
+      });
+    });
+
+    it('emits POP with current location key', function(done) {
+      // set initial state, this is needed because all implementations gets state from different places
+      history.pushState({ initial: 'state' }, '/');
+
+      // now create history for testing if initial POP event has location.key
+      history = createHistory();
+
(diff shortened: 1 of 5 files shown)


michalkvasnicak:
@mjackson what do you think about this?

modules/__tests__/describeInitialLocation.js
```diff
@@ -0,0 +1,60 @@
...
+      history = createHistory();
+
+      unlisten = history.listen(function(location) {
+        try {
+          expect(location.action).toEqual(POP);
+          expect(location.key).toNotEqual(null);
+
+          // in case of memory history we can get saved state, because generated key is always unique
+          // and even if we save state to sessionStorage, we can not get it between instances of MemoryHistory
+          // because key will be not mapped to given state
```
mjackson: Maybe we should make MemoryHistory work like the others and use a module-level storage (instead of per-instance) so we don't need this check?

modules/createHistory.js
```diff
@@ -49,7 +54,9 @@ function createHistory(options={}) {
     if (location) {
       listener(location);
     } else {
-      updateLocation(getCurrentLocation());
+      var locationReplacement = getCurrentLocation();
+      replaceState(locationReplacement.state, locationReplacement.pathname + locationReplacement.search);
```
mjackson: I think I'd like to avoid using `replaceState` here if we can. As it is currently implemented, `replaceState` will run through `confirmTransition`, which I think we want to avoid on the initial page load. Also, BrowserHistory does a full page refresh in browsers that don't support HTML5 history when you call `replaceState`, so it will leave those users in an infinite refresh cycle. Instead, I think we can simply update the key of a history inside each implementation's `getCurrentLocation` function.

mjackson:
Sorry for the long feedback cycle @michalkvasnicak. I've been heads down trying to get React Router v 1.0 out the door. Hopefully the feedback I've given here can be useful.

Also, I think it's probably best if we leave changes to the auto-generated `lib` files out of PRs and only regenerate them when we do a release. Thanks!

michalkvasnicak:
@mjackson There is problem that `createKey` is exposed after `createHistory()` is called. If we want to generate key for location in `getCurrentLocation` of given `create*History` it won't be accessible because we need to pass this function to `createHistory` options.

Any idea? Maybe extract `createRandomKey` to `utils.js` and not provide possibility to change `keyLength` or in userland?

Issue #15, opened 2015-07-29
Support return false in transition hooks


mjackson:
`return false` in a transition hook cancels the transition.

RobertKielty:
Hi Michael, I can take a look at this now.

mjackson:
@RobertKielty Thank you!

You'll probably need to do a bit of refactoring [here](https://github.com/rackt/history/blob/f09171b73067ae042a0042efdec8b4ad1668fe66/modules/createHistory.js#L69-L88).

ericlynnpeterson:
Etiquette question:
When someone mentions they are taking a look at an issue, is it impolite to also work on a solution?

ryanolsonx:
I should have checked if someone was planning to work on this issue :/  That was bad etiquette on my part.  I'm going to just close my pull request. Sorry about that!

RobertKielty:
@ericlynnpeterson no, I don't think so, as long as the job gets done; Good on ya @ryanolsonx ! @mjackson are you happy with Ryan's change?

ericlynnpeterson:
I think the change looks good, too.  I had missed the pull request when I started working on the issue myself and I just wanted to clarify.  :-)

RobertKielty:
We're fighting over the sweeping brush :)

ryanolsonx:
Alright, well.. I'm just cleaning up some of my commits and I'll re-submit my pull request and if @mjackson likes it, he can merge it :)

RobertKielty:
(thumbs up)

mjackson:
Thanks, everyone! Let's follow up in #19.

PR #19, opened 2015-08-05
Transition hook that returns false will cancel transition (Issue #15)

ryanolsonx:
This change adds support to allow a developer to `return false` from their transition hook to cancel the transition.

Thanks,

Ryan

### DIFF
--- a/modules/createHistory.js
+++ b/modules/createHistory.js
@@ -69,7 +69,7 @@ function createHistory(options={}) {
   function getTransitionConfirmationMessage() {
     var message = null;
 
-    for (var i = 0, len = transitionHooks.length; i < len && message === null; ++i)
+    for (var i = 0, len = transitionHooks.length; i < len && message == null; ++i)
       message = transitionHooks[i].call(this);
 
     return message;
@@ -78,7 +78,7 @@ function createHistory(options={}) {
   function confirmTransition(callback) {
     var message = getTransitionConfirmationMessage();
 
-    if (getUserConfirmation && message) {
+    if (getUserConfirmation && typeof message === 'string') {
       getUserConfirmation(message, function (ok) {
         callback(ok !== false);
       });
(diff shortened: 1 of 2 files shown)


modules/createHistory.js
```diff
@@ -69,21 +69,21 @@ function createHistory(options={}) {
   function getTransitionConfirmationMessage() {
     var message = null;
 
-    for (var i = 0, len = transitionHooks.length; i < len && typeof message !== 'string'; ++i)
+    for (var i = 0, len = transitionHooks.length; i < len && message === null; ++i)
```
mjackson: Using `message === null` here will break the loop after the very first function that returns `undefined`, which is any function that doesn't have a `return` statement. Let's use `message == null` instead.

modules/createHistory.js
```diff
@@ -69,21 +69,21 @@ function createHistory(options={}) {
...
 
     return message;
   }
 
   function confirmTransition(callback) {
-    var message;
+    var message = getTransitionConfirmationMessage();
 
-    if (getUserConfirmation && (message = getTransitionConfirmationMessage())) {
+    if (getUserConfirmation && message) {
```
mjackson: Let's use `typeof message === 'string'` here so we don't pass non-strings to `getUserConfirmation`.

modules/createHistory.js
```diff
@@ -69,21 +69,21 @@ function createHistory(options={}) {
...
       getUserConfirmation(message, function (ok) {
         callback(ok !== false);
       });
     } else {
-      callback(true);
+      callback(message !== false);
```
mjackson: Solid. :+1:

mjackson:
Thank you @ryanolsonx! Don't worry about the failing tests on Travis. I'm working on getting that sorted out with the BrowserStack people. The tests should all pass locally tho.

ryanolsonx:
No problem! I'll make the suggested changes right now

mjackson:
Thanks, @ryanolsonx!

ryanolsonx:
Awesome, thanks @mjackson! I'm hoping to contribute more to this project and/or react-router in the future. 👍

Issue #22, opened 2015-08-10
Transition Middleware


aldendaniels:
I'd like to hook into history to do data loading. To do this, I need to be able to pause a transition while data loading is occurring. 

I think this can be simulated today using `registerTransitionHook()` and a custom `getUserConfirmation()` handler that doesn't actually show a user confirmation dialog. This is of course a hack.

A better solution would look something like this:

```javascript
let history = createBrowserHistory();
history.registerTransitionMiddleware(function(nextLocation, prevLocation, fnNext, fnAbort) {
   fetchSomeData(location).then(fnNext);
});
```

The transition would simply freeze until `fnNext` or `fnAbort` are called.

If the direction sounds good, then I could submit a PR for this.

mjackson:
> I need to be able to pause a transition while data loading is occurring

I don't think you'd actually want to do this. For the user, the page would be non-responsive while you're fetching data (i.e. they would click the back button and nothing would happen for a while).

aldendaniels:
@mjackson - this simulates the browser's native behavior - if I click a link on a page, then I'm left looking at the current page until the target page loads. 

You would, of course, want to show a loading indicator during the transition phase, to keep the UI from feeling unresponsive.

gaearon:
@aldendaniels 

We used this approach and it turned out to be a disaster. For example, while data is loading, what does Back button do? Does it cancel transition? If not, how can you cancel a transition that's taking too long (e.g. bad mobile connection). So does Back cancel pending transition? But that looks very weird on mobile, where "Back" is a slide gesture.

My advice: change URL right away. Changing it later = bad UX. I've learned it hard way.

aldendaniels:
@gaearon - Thanks for weighing in.

The browser natively does exactly what you're describing as being problematic on mobile - if you slide to go back during a pending transition then you slide "back" onto the same page you were on (cancelling the transition). Since this is native behavior, why not simulate it?

Are there other complications I'm missing for an SPA?

mjackson:
What don't you like about doing something like this, @aldendaniels?

```js
history.listen(function (location) {
  fetchTheData(location, function (data) {
    renderThePage(location, data);
  });
});
```

The current implementation of transition hooks is minimal on purpose. It only addresses the specific use case of **confirming transitions away from a route**. If you need to run some logic **after** a user hits a route, you can already easily do that.

aldendaniels:
There are a couple of issues: 

[...]

PR #23, opened 2015-08-11
Add transition middleware

aldendaniels:
#### Status
_WIP_: Working, but needs more tests. Making PR now to get feedback.

#### Overview
This is a PR for the functionality described in #22. This lets users register history middleware that can delay, abort, or redirect transitions programatically. A primary use-case for this is to perform data loading or authentication verification before transitioning to a new route.

The new API adds two new methods:

 - `registerTransitionMiddleware()`
 - `unregisterTransitionMiddleware()`

#### Example
```javascript
function handler(nextLocation, prevLocation, fnNext) {
  fnNext(); // continue;
  // OR
  fnNext(false); // abort
  // OR
  fnNext(createLocation(...)) // redirect.
}

// Register middleware.
let unregister = history.registerTransitionMiddleware(handler);

// Unregister middleware.
unregister() // OR...
history.unregisterTransitionMiddleware(handler); 
```

#### Implementation Details

The middleware will be invoked
 1. When the history object is initially created
 2. On every subsequent transition

Because transitions can now be async, we no longer assert when a new transition interrupts an existing _pending_ transition. Instead, we just silently ignore the interrupted transition.

#### Possible Deprecation
Middleware could replace:

 - the `registerTransitionHook()` method
 - the `unregisterTransitionHook()` method
 - the `getUserConfirmation` option

The functionality to prompt the user can be replaced with:

```javascript
history.registerTransitionMiddleware((ignore1, ignore2, fnNext) => {
  fnNext(confirm('Are you sure?'));
});
```


aldendaniels:
Right now the middleware is run concurrently, which isn't right. I'll make an update shortly to run the middleware serially in the registered order.

I'd also like for middleware to be able to redirect a transition, instead of just aborting the transition. I need to give a bit more thought to how this should work.

aldendaniels:
Looks like CI tests are failing due to missing BrowserStack credentials.

mjackson:
Hey @aldendaniels! Thanks for submitting this PR. Let's discuss in #22 before moving forward here.

mjackson:
Closing based on the conversation in #22.

Issue #17, opened 2015-08-05
Add scroll position support


mjackson:
We need to add a `getScrollPosition` option to history objects so we can store the scroll position before navigating to a new page. Location objects should have this data in the `scrollX` and `scrollY` properties.

```js
var history = createHistory({
  getScrollPosition() {
    return { x: 0, y: 0 };
  }
});

history.listen(function (location) {
  console.log(location.scrollX, location.scrollY);
});
```

ericlynnpeterson:
I'm trying to wrap my head around this project, so thanks in advance for being patient.

Is this issue saying that
  1) We want to know the scroll position for the *previous* location object?
  2) We want to provide the initial scroll position for the *next* location object?
  3) Both?

mjackson:
Hi @ericlynnpeterson! Thank you for taking a look :)

It may help to think about it like this: The `location` object tells you at any given point in time *where you are*. Currently, the pathname, search string, and state are all part of that object. They all answer the question "where am I?"

All we want to do here is add scroll information to the `location` object so it can better answer that question. This will most likely involve two parts:

1. Use `getScrollPosition` to record the scroll information immediately before we change the URL
2. Include the `scrollX` and `scrollY` properties in the `createLocation` function

As a side note, `createLocation` is getting quite a few arguments. We may want to consider slimming them down into an options object or something...

Does that help?

ericlynnpeterson:
Thanks @mjackson.  That does help.  Maybe a followup question?

* If the `location` object answers the questions "where am I?", with regards to scrolling, do we need to update the current `location` object's scrollX and scrollY position on every scroll?

Maybe an example will help.
```js
// Initial Location with scroll position

var history = createHistory({
  getScrollPosition() {
    return { x: 0, y: 200 };
  }
});

// Initial location
{
  pathname: '/',
  search: '',
  state: null,
  action: POP,
  scrollX: 0,
  scrollY: 200 // <-- from getScrollPosition()
}

// Scroll down a bit; updates along the way
{
  pathname: '/',
  search: '',
  state: null,
  action: POP,
  scrollX: 0,
  scrollY: 400
}

// Push a new location
history.pushState(null, '/home');
{
  pathname: '/home',
  search: '',
  state: null,
  action: PUSH,
  scrollX: 0,
  scrollY: 0
  // I was thinking that the only time we used the getScrollPosition()
  // value was for the initial location.  Correct?
}

// Scroll down in the new location a bit; updates along the way
{
  pathname: '/home',
[...]

PR #24, opened 2015-08-11
Add getScrollPosition to History, and scrollX/scrollY to Location

duncanmak:
I just typed up the code snippets from @mjackson and @gaearon in the comments above, and added just a little bit of error handling (there's now a fallback if `getScrollPosition` is not supplied).

I'm also looking into writing a test for this, I'm stuck because I don't know how to mock the scroll position changing in a test.

Also, I don't know if I should include the generated changes in `lib/umd/History.js` and `lib/umd/History.min.js`.

This addresses Issue #17.

### DIFF
--- a/modules/createDOMHistory.js
+++ b/modules/createDOMHistory.js
@@ -24,13 +24,18 @@ function startBeforeUnloadListener({ getTransitionConfirmationMessage }) {
   };
 }
 
+function getScrollPosition() {
+  return { x: window.scrollX, y: window.scrollY };
+}
+
 function createDOMHistory(options) {
   var history = createHistory({
     getUserConfirmation,
     ...options,
     saveState,
     readState,
-    go
+    go,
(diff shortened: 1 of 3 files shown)


mjackson:
Thanks for this PR @duncanmak! I think I may not have explained myself very well in #17.

This PR adds the scroll position to the **next** location, but we need to store scroll position on the **previous** location so that when we go **back** to it we'll know where to scroll the page. In order to persist state across transitions, we'll need to store some stuff in `location.state`, probably under a reserved key.

I'll take a stab at this today and see if I can show you what I mean. Code may be more clear than words :)

mjackson:
I think we can get by using the `setState` method introduced in 63091d2c6ce15a8a9a4d188923e86607dcbf9615 instead of `getScrollPosition`. Closing for now.

Thanks again, @duncanmak!

PR #28, opened 2015-08-16
Add location.current and improve cancelling POP transitions

taurose:
Builds upon `allKeys` from https://github.com/rackt/history/commit/96d3b2c33590e6359c3f1393946f223bd7325122 to provide `location.current`. Also fixes two things:

1. key from initial location wasn't put into `allKeys`
2. `allKeys` wasn't modified correctly for push and replace, not taking the current position into account (e.g. push should remove all keys that come after the current one)

Possible future work:

* persist keys in `sessionStorage` so it works after refresh. This would require building multiple `allKeys` lists to support multiple browsing sessions
* figure out edge cases related to using hash navigation with BrowserHistory, and try to fix or document them

PR #29, opened 2015-08-17
Improve cancelling POP transitions

taurose:
Follow-up of #28.

Fixes:

1. key from initial location wasn't put into `allKeys`
2. `allKeys` wasn't modified correctly for push and replace, not taking the current position into account (e.g. push should remove all keys that come after the current one)

### DIFF
--- a/modules/createHistory.js
+++ b/modules/createHistory.js
@@ -29,13 +29,25 @@ function createHistory(options={}) {
 
   var allKeys = [];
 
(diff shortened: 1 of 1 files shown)


mjackson:
Thanks, @taurose!

pke:
This still doesn't seem to work properly:
https://github.com/reactjs/react-router/issues/3738

PR #30, opened 2015-08-17
Convert enableQueries to higher-order function (like a decorator)

acdlite:
Extensions like `enableQueries` are more powerful if they operate on a function (`createHistory`) rather than the thing it returns (`history`). The reason in a nutshell is that it allows the extension to intercept/modify both the input and output of the function. For example, this is how decorators work. It's also how store enhancers, like `applyMiddleware()` and `devtools()`, work in Redux.

For simple extensions like enableQueries, it doesn't end up making a huge difference, but for future extensions it may. Better to establish the convention now rather than later :)

(Because of indentation, this looks like a bigger PR than it actually is. The key lines that have changed are 11-13 of `enableQueries.js`.)

### DIFF
--- a/modules/__tests__/describeQueries.js
+++ b/modules/__tests__/describeQueries.js
@@ -11,7 +11,7 @@ function describeQueries(createHistory) {
   describe('query serialization', function () {
     var history, unlisten;
     beforeEach(function () {
-      history = enableQueries(createHistory());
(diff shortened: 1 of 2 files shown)


mjackson:
Looks great, and sound reasoning. Thank you!

PR #33, opened 2015-08-20
Expose createPath

acdlite:
This makes `createPath()` part of the standard history interface. Previously it was only added by `enableQueries()`, but I think it's useful for any consumer to be able to call `createPath()`, regardless of whether queries are enabled or not.


mjackson:
Thanks, @acdlite!

PR #40, opened 2015-09-07
Fix window.history.replaceState

vslinko:


vslinko:
Second argument of `replaceState` is `title` not `url`.

mjackson:
Whoa, what an oversight! Thanks, @vslinko!

PR #41, opened 2015-09-08
Fix MemoryHistory "top of stack" check

jinroh:
`current` has only been incremented at the time of the check.


mjackson:
Can you please add a test for this @jinroh?

jinroh:
@mjackson done.

mjackson:
Thanks, @jinroh!

PR #43, opened 2015-09-11
Don't create a new history entry when transitioning to the current path

taurose:
This PR normalizes the behavior of all histories to not create a new history entry when pushing the path that's already active. Like browsers do with static content.

Currently, all histories create new entries, except `HashHistory` without `queryKey` since it's unable to do so. Instead, it'll print a [warning](https://github.com/rackt/history/blob/master/modules/createHashHistory.js#L103) which doesn't seem right considering it's a common scenario.

I implemented it in the history's `finishTransition` methods to treat these cases like a `REPLACE`. `location.action` will still be `PUSH` though. Alternatively, I think this could be implemented in the base history to change `location.action` to `REPLACE`. That's what I was doing in an older [PR](https://github.com/rackt/react-router/blob/836032130c4cffa7b6d97b892043f1d7ac2cd1f3/modules/History.js#L171).

Also related: https://github.com/rackt/react-router/pull/1031

### DIFF
--- a/modules/createHistory.js
+++ b/modules/createHistory.js
@@ -114,6 +114,16 @@ function createHistory(options={}) {
         return // Transition was interrupted.
 
       if (ok) {
+        // treat PUSH to current path like REPLACE to be consistent with browsers
+        if (nextLocation.action === PUSH) {
+          let { pathname, search } = getCurrentLocation()
+          let currentPath = pathname + search
+          let path = nextLocation.pathname + nextLocation.search
+
+          if (currentPath === path)
+            nextLocation.action = REPLACE
+        }
+
         if (finishTransition(nextLocation) !== false)
           updateLocation(nextLocation)
       } else if (location && nextLocation.action === POP) {
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,9 @@
 ## HEAD
 
 - Fail gracefully when Safari security settings prevent access to window.sessionStorage
+- Pushing the currently active path will result in a replace to not create additional browser history entries (see [#43])
+
+[#43]: https://github.com/rackt/history/pull/43
 
 ## [v1.13.0]
 > Oct 28, 2015
--- a/modules/__tests__/describePushState.js
+++ b/modules/__tests__/describePushState.js
@@ -1,6 +1,6 @@
 /*eslint-env mocha */
 import expect from 'expect'
-import { PUSH, POP } from '../Actions'
+import { PUSH, POP, REPLACE } from '../Actions'
 import execSteps from './execSteps'
 
 function describePushState(createHistory) {
@@ -30,6 +30,35 @@ function describePushState(createHistory) {
           expect(location.search).toEqual('?the=query')
           expect(location.state).toEqual({ the: 'state' })
           expect(location.action).toEqual(PUSH)
+        }
+      ]
+
+      unlisten = history.listen(execSteps(steps, done))
+    })
+
+    it('becomes a REPLACE if path is unchanged', function (done) {
+      let steps = [
+        function (location) {
+          expect(location.pathname).toEqual('/')
+          expect(location.search).toEqual('')
+          expect(location.state).toEqual(null)
+          expect(location.action).toEqual(POP)
+
+          history.pushState({ the: 'state' }, '/home?the=query')
+        },
+        function (location) {
+          expect(location.pathname).toEqual('/home')
+          expect(location.search).toEqual('?the=query')
+          expect(location.state).toEqual({ the: 'state' })
+          expect(location.action).toEqual(PUSH)
+
+          history.pushState({ different: 'state' }, '/home?the=query')
+        },
+        function (location) {
+          expect(location.pathname).toEqual('/home')
+          expect(location.search).toEqual('?the=query')
+          expect(location.state).toEqual({ different: 'state' })
+          expect(location.action).toEqual(REPLACE)
         }
       ]
 
--- a/modules/createHashHistory.js
+++ b/modules/createHashHistory.js
@@ -95,28 +95,27 @@ function createHashHistory(options={}) {
 
     let path = (basename || '') + pathname + search
 
-    if (queryKey)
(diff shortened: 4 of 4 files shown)


mjackson:
> Alternatively, I think this could be implemented in the base history to change location.action to REPLACE

This feels more semantically correct to me. Just because `pushState` was called doesn't mean we necessarily need to emit a push `location`. If we're really just doing a replace behind the scenes, we should probably let the user know.

taurose:
You're probably right. I was thinking with `PUSH` one might still be able to figure out what happened if we add `location.current` or `location.delta`.

I changed the implementation. Let me know if/how you want this tested.

ricardosoeiro:
Is this PR going to be included? I think it makes more sense than the current default...

mjackson:
Yes, I'd definitely like to move forward here. Just haven't had time to take a look!

@taurose I'm ok with whatever tests you'd like to write here. Maybe we could write a test that only tests hash history and does a `pushState` and then checks to make sure `location.action` is actually a REPLACE?

Sorry for taking so long to come back around here.

th0r:
Guys, maybe add the `REFRESH` action?

In my project I've made a `useRefresh` middleware that adds `history.refresh()` method that calls `history.listen` handler with the current location, but sets it's action to `REFRESH`.

I need it to softly "reload" the page (refetch all the routes data) and it would be great to be the part of the core.

taurose:
Just rebased and added a test that checks if `pushState` to the same path will result in a REPLACE action.

@mjackson: Why only test hash history? That behavior affects all histories. I had to do some changes to hash history to not print a warning, but I don't think it makes sense to test this. 

@th0r: This wouldn't re-execute react-router transition hooks, would it? Either way, I think this should be a separate feature request.

ricardosoeiro:
Looking good :)

taion:
@taurose @mjackson Ping! Should we rebase and merge this?

ricardosoeiro:
Still waiting for this PR... how can I help, to move this forward?

taurose:
No objections here. I was hoping for @mjackson or someone else to take another look :)

NeXTs:
+1 waiting for this PR

mjackson:
I'm taking a look at this now, @taurose. Thanks for your patience :)

modules/createHistory.js
```diff
@@ -114,6 +114,16 @@ function createHistory(options={}) {
         return // Transition was interrupted.
 
       if (ok) {
+        // treat PUSH to current path like REPLACE to be consistent with browsers
+        if (nextLocation.action === PUSH) {
+          let { pathname, search } = getCurrentLocation()
```
mjackson: I'm not 100% sure this is the right place to do this, but should be ok for now. My concern is for environments where we may not be able to synchronously call `getCurrentLocation`, like native, where we have to use `AsyncStorage` to store everything.

modules/createHistory.js
```diff
@@ -114,6 +114,16 @@ function createHistory(options={}) {
...
+          let currentPath = pathname + search
+          let path = nextLocation.pathname + nextLocation.search
```
taurose: @mjackson I just realized that I haven't considered `location.hash`. Seems like the two lines here should use `createPath(location)`, right?

modules/createHistory.js
taurose: I think in that case we'd want to hold on to the current location.

joshacheson:
Any chance on a timeline for this getting published to npm? Using `history` in production, and this fix will be a huge help in solving some problems i've been having caused by `redux-router` / `react-router` and hash histories

taion:
The npm changes were merged. There's nothing that needs to be changed in this library. React Router's rc3, rc4, and final builds all specifically exclude `history` through a separate workaround anyway.

Other packages will see this fix as they get re-published with the new version of npm.

just-boris:
@taion well, when these changes will be published on npm? It is so hacky download package directly from github

taion:
That's really up to the maintainers of each of the packages that you use.

just-boris:
@taurose now I am about the `history` package. Or this question is not for you?

taion:
Oops, disregard what I said earlier - I got confused about which issue I was on.

taion:
Yeah it would be pretty nice to get this released on npm when possible.

mjackson:
Just released in version 1.13.1. Enjoy.

NeXTs:
@taurose Thank you so much! No more annoying messages at console with redux-router and hash history

just-boris:
Tested on my app too. Working fine, thanks!

phuson:
I have a case where I have a bunch of anchors on a page, and want to record when a user navigates between those anchors.  With the change in this PR, it removes that functionality, because now it treats all hash changes on the same page as if they're one single item in the browser history.  Is there a way to optionally allow the old functionality for those of us who need it?  Alternately, is there a way to force a `pushState` for items we want to maintain in the history, instead of automatically replacing the states if the URLs are considered to be the same?

taion:
@phuson 

That doesn't sound good. Seems like we should be checking the full computed path instead.

Would you mind opening this as a new issue? It looks like a regression introduced here, and it'd be better to track it separately.

modules/createHistory.js
taion: Why is it correct to use `getCurrentLocation` here rather than just `location`?

modules/createHistory.js
taion: I think `createPath` is the correct check here. It doesn't account for `basename` (and that part is a bit messy honestly in `createBrowserHistory`), but `basename` isn't going to change on us anyway, so it doesn't matter. That would also fix https://github.com/rackt/history/pull/43#issuecomment-160782080.

taion:
@taurose Do you think you'll have a chance to address the issues? If not, I can amend the code in question.

phuson:
@taion Thanks for looking into this! Here's the new issue for tracking purpose: https://github.com/rackt/history/issues/166

taion:
@taurose Never mind, I put together a PR with the fix (:

PR #44, opened 2015-09-11
Check that history is supported before trying to replaceState

AndreasHogstrom:
All other calls to replaceState et. al. does a check of isSupported to make sure it will work, except this one. I believe is the correct solution, at least it solves my IE problems!


pucka:
I really hope this pr will be merged. Without it [this docs are incorrect](https://github.com/rackt/react-router/blob/master/docs/guides/basics/Histories.md#ie8-ie9-support)

Furizaa:
+1

PR #51, opened 2015-09-14
don't strip the hash off of the window path

jeffreywescott:
Without this change, when using `createHistory` or `createBrowserHistory`, the `window.location.hash` gets stripped off of the location, never to be recovered, breaking in-page navigation when using react-router.

### DIFF
--- a/modules/DOMUtils.js
+++ b/modules/DOMUtils.js
@@ -27,7 +27,7 @@ export function replaceHashPath(path) {
 }
 
 export function getWindowPath() {
-  return window.location.pathname + window.location.search
+  return window.location.pathname + window.location.search + window.location.hash
 }
 
 export function go(n) {
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
   },
   "bugs": "https://github.com/rackt/history/issues",
   "scripts": {
-    "build": "babel ./modules --babelrc .babelrc -d lib --ignore '__tests__'",
+    "build": "babel ./modules --stage 0 --loose -d lib --ignore '__tests__'",
     "build-umd": "NODE_ENV=production webpack modules/index.js umd/History.js",
     "build-min": "NODE_ENV=production webpack -p modules/index.js umd/History.min.js",
     "start": "webpack-dev-server -d --content-base ./ --history-api-fallback --inline modules/index.js",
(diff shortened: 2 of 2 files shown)


ryancbarry:
:+1:

JacobOtte:
:+1:  This is important stuff.

jackmoore:
This isn't fixed in 1.11.0.  Using history.pushState with a hash strips the hash. Using history.pushState with a hash and a query strips both.

`history.pushState(null, '/home#top')` redirects to `/home`

`history.pushState(null, '/home#top', {page: 2})` also redirects to `/home`

janpieterz:
@mjackson are you aware of this? I'm not a 100% sure but it seems like it indeed still doesn't work (using latest history combined with react-router).

zdavis:
Would be great to see this merged in.

zdavis:
Yes, seems to not be fixed in 1.12.1. I believe the problem is here:

https://github.com/rackt/history/blob/master/modules/createBrowserHistory.js#L70

...Location is being rebuilt without the hash

I'll try to put together a pull request.

mjackson:
Thanks for bringing this to my attention, @jackmoore. I think @zdavis is right about where the problem is. I'll see if I can make a fix.

mjackson:
I should also note that in your second example, @jackmoore, you should be sure you're using the `useQueries` enhancer. That will give you a 3rd `query` arg to `pushState`.

mjackson:
@jackmoore @janpieterz @zdavis This is fixed in 7012f9b697f614c1d412926b04079d6b86732757

mjackson:
Fix released in 1.12.2

janpieterz:
Awesome, thanks a lot for the quick effort!

jackmoore:
@mjackson Thanks, but when I combine the hash with query, it's putting the query in the hash.  I'm using react-router's Link component if that matters.  Am I doing something wrong here?

```
<Link to='/home#top' query={{test: true}}>Test</Link>
```

Redirects to: `/home#top?test=true`

zdavis:
Thanks so much!

mjackson:
@jackmoore That's a slightly separate issue. Let's follow up in https://github.com/rackt/react-router/issues/2176

PR #59, opened 2015-09-16
Add support to async history update

esnunes:
In order to use react-router together with react-blessed the memory-based history must support async call of listeners.

By default the memory-based history has a sync nature which causes an unexpected behaviour during mount of react components due to a history update that triggers a state change on Router component that was not initialized (https://github.com/rackt/react-router/blob/v1.0.0-rc1/modules/Router.js#L57-L63).

Error message:
```bash
TypeError: Cannot read property '_currentElement' of null
    at [object Object].ReactCompositeComponentMixin._updateRenderedComponent (/Users/nunes/Developer/js/sqlectron/sqlectron-term/node_modules/react/lib/ReactCompositeComponent.js:551:52)
    at [object Object].ReactCompositeComponentMixin._performComponentUpdate (/Users/nunes/Developer/js/sqlectron/sqlectron-term/node_modules/react/lib/ReactCompositeComponent.js:536:10)
    at [object Object].ReactCompositeComponentMixin.updateComponent (/Users/nunes/Developer/js/sqlectron/sqlectron-term/node_modules/react/lib/ReactCompositeComponent.js:465:12)
    at [object Object].wrapper [as updateComponent] (/Users/nunes/Developer/js/sqlectron/sqlectron-term/node_modules/react/lib/ReactPerf.js:66:21)
    at [object Object].ReactCompositeComponentMixin.performUpdateIfNecessary (/Users/nunes/Developer/js/sqlectron/sqlectron-term/node_modules/react/lib/ReactCompositeComponent.js:413:12)
    at Object.ReactReconciler.performUpdateIfNecessary (/Users/nunes/Developer/js/sqlectron/sqlectron-term/node_modules/react/lib/ReactReconciler.js:102:22)
    at runBatchedUpdates (/Users/nunes/Developer/js/sqlectron/sqlectron-term/node_modules/react/lib/ReactUpdates.js:129:21)
    at ReactBlessedReconcileTransaction.Mixin.perform (/Users/nunes/Developer/js/sqlectron/sqlectron-term/node_modules/react/lib/Transaction.js:136:20)
    at ReactUpdatesFlushTransaction.Mixin.perform (/Users/nunes/Developer/js/sqlectron/sqlectron-term/node_modules/react/lib/Transaction.js:136:20)
    at ReactUpdatesFlushTransaction.assign.perform (/Users/nunes/Developer/js/sqlectron/sqlectron-term/node_modules/react/lib/ReactUpdates.js:86:38)
```

### DIFF
--- a/modules/createMemoryHistory.js
+++ b/modules/createMemoryHistory.js
@@ -27,6 +27,13 @@ function createMemoryHistory(options={}) {
     go
   })
 
+  if (options.asyncListen) {
+    const syncListen = history.listen;
+    history.listen = listener => {
+      setTimeout(syncListen.bind(history, listener), 0);
+    }
+  }
+
   let { entries, current } = options
 
   if (typeof entries === 'string') {


gaearon:
Isn't this `react-blessed` issue then?
Either way, I'd rather see this in userland, e.g. `history-listen-async`.

```
export default function listenAsync(createHistory) {
  return (...args) => {
    let history = createHistory(...args);
    let originalListen = history.listen.bind(history);
    history.listen = (listener) => originalListen((...listenerArgs) => {
      setTimeout(() => listener(...listenerArgs), 0);
    });
    return history;
  };
}
```

```js
import listenAsync from 'history-listen-async';
import createHistory from 'history/lib/createHashHistory';

let history = listenAsync(createHistory)({ options });
```

esnunes:
Hi @gaearon,

I've only identified this problem using it connected to react-router and redux. I believe it might be something related to the usage of "empty" (render returning false) components and react-blessed. It might be something wrong with react-blessed implementation too.

In the meanwhile I'm using a very similar approach you've mentioned however I decided to create this pull request due to the fact that I believe the major usage of this history component is related to react-router and specially the browser history implementation which by its nature works in async mode. Wouldn't be right that all implementations behave in the same way?

taurose:
Don't the other histories call the listen callback synchronously as well? I think the only discrepancy is `go(n)` and related, which is sync for memory history, but async for others.

gaearon:
@esnunes What version of react-blessed are you using?

esnunes:
You are right @taurose, I thought the browser version was triggering listener on window.history update.

@gaearon I'm using the latest version of react-router, react-blessed and redux. You can check the source code here: https://github.com/sqlectron/sqlectron-term . In case you want to run it without async listen it is just replace the `createHistory` method on `lib/index.js` by the createMemoryHistory. Suggestions are welcome!

mjackson:
@esnunes Can't you just do this in your listener?

```js
history.listen((location) => {
  setTimeout(() => asyncListener(location), 0)
})
```

No need to use a flag here AFAICT.

esnunes:
Hi @mjackson ,

I followed @gaearon suggestion and I've created my [own history](https://github.com/sqlectron/sqlectron-term/tree/master/src/history).

mjackson:
Excellent. Closing.

Issue #68, opened 2015-09-25
useBasename results in empty location when matches exactly


jesenko:
When using `useBasename` together with `react-router`, initial location that matches `basename` exactly results in `location=''` which cannot be matched by `react-router` to any route. E.g:

```jsx
<Route component={App}>
  <Route path="/" component={MyComponent} />
</Route>
...
const history = useBasename(createBrowserHistory)({
  basename: '/basename'
});
```
Entering URL `/basename` in browser address bar results in 
`Warning: Location "" did not match any routes`

Not sure whether issue is with `react-router` not matching empty location "/" or with useBasename, mishandling this specific case... Problem can be fixed on `useBasename` side by simply replacing empty pathname with `/`

PR #69, opened 2015-09-25
Set pathname to '/' when stripped pathname is empty

jesenko:
Related to #68.


mjackson:
Excellent catch. Thanks, @jesenko!

PR #75, opened 2015-09-30
Catch `QuotaExceededError`

### LINKED ISSUE #42: QuotaExceededError in Safari incognito mode
When using safari in incognito mode, navigation fails with the following exception:

```
QuotaExceededError: DOM Exception 22: An attempt was made to add something to storage that exceeded the quota.
```

Safari denies usage of local/session storages when in incognito mode, and raises this error when you try to.

I am using the BrowserHistory history and it seems like the error stems from history/DOMStateStorage.

iam4x:
fixes #42

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 lib
 umd
+node_modules
--- a/modules/DOMStateStorage.js
+++ b/modules/DOMStateStorage.js
@@ -1,3 +1,5 @@
+import warning from 'warning'
+
 /*eslint-disable no-empty */
 const KeyPrefix = '@@History/'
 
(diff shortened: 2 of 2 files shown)


mjackson:
Thanks for the PR, @iam4x. I'd like to address both parts of the solution from https://github.com/rackt/history/issues/42#issuecomment-139453081 in the same PR. Are you up for it?

iam4x:
Sure, but I'll need to understand the whole library before or some hints ;)

.gitignore
```diff
@@ -1,2 +1,3 @@
 lib
 umd
+node_modules
```
mjackson: Please remove this and put it in your `~/.gitignore` instead.

mjackson:
This looks good. I can merge and take it from here.

.gitignore
taion: This actually didn't get taken out, but I think it's fine to keep - it's part of standard `.gitignore` for Node per e.g. https://github.com/github/gitignore/blob/master/Node.gitignore. Most other JS projects ignore `node_modules`, per e.g. https://github.com/rackt/redux/blob/v3.0.4/.gitignore#L3.

Typically I think the user-global `.gitignore` is better suited for editor configuration rather than anything even slightly project-specific.

JitendraThakur007: testing

PR #79, opened 2015-10-05
Make createLocation an instance method

mjackson:
A single, global createLocation method cannot be aware of history enhancers like `useQueries`. So users need to manually modify location objects when creating them programmatically instead of getting them in a `listen` callback.

This commit makes `createLocation` an instance method of history objects and deprecates the `createLocation` method on the top-level exports.

### DIFF
--- a/modules/createLocation.js
+++ b/modules/createLocation.js
@@ -1,4 +1,5 @@
 import warning from 'warning'
+import deprecate from './deprecate'
 import { POP } from './Actions'
 
 function extractPath(string) {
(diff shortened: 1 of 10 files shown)


mjackson:
This should fix https://github.com/rackt/react-router/issues/2015

markdalgleish:
That was fast! Nice work :smile:  :+1:

modules/createLocation.js
```diff
@@ -48,4 +49,7 @@ function createLocation(path='/', state=null, action=POP, key=null) {
   }
 }
 
-export default createLocation
+export default deprecate(
+  createLocation,
+  'createLocation is deprecated; use history.createLocation instead'
```
robcolburn: This error message is confusing… this still sounds valid.
```js
import {createLocation} from 'history';
```

mjackson: Let's discuss in a PR with your improvements :)

robcolburn: #82

PR #92, opened 2015-10-09
Don't ship all the build scripts

taion:
Same as https://github.com/rackt/react-router/pull/2222


mjackson:
Ah! Sorry. I should've waited a bit :) 82de9767a2f69d5de5b1b171813fdb32d5956f4a

PR #98, opened 2015-10-12
Proposal: Allow .listen(cb, false) to not be called immediately

jamischarles:
This is a proposal. If you're onboard with this API change I'll gladly do all the leg work to get everything working / tested / documented etc...

**Here's my use case**:
I am opening a modal. If I use back/forward buttons I want the modal to close.

For this use case, the callback being called immediately causes problems.

I want to unbind when:
- modal closes
- url navigates (listenToOnce essentially)

**What I currently have to do**:
```js
let unlistenToHistory;

{
	onModalOpen: function() {
		unlistenToHistory = history.listen((location) => {
			if (unlistenToHistory){
				// unbind the urlChange listener
				unlistenToHistory();
				this.closeModal();
			}
		});
	},
	onModalClose: function() {
		if (unlistenToHistory) {
			unlistenToHistory(); // unbind the urlChange listener
			unlistenToHistory = undefined;
		}
	},
}
```

**What I'd love, is to be able to do this**:
```js
let unlistenToHistory;

{
	onModalOpen: function() {
		unlistenToHistory = history.listen((location) => {
			unlistenToHistory(); // this will be set, since cb isn't called immediately
			this.closeModal();
		}, false);
	},
	onModalClose: function() {
		if (unlistenToHistory) {
			unlistenToHistory(); // unbind the urlChange listener
			unlistenToHistory = undefined;
		}
	},
}
```

TODO:
- [ ] Add tests
- [ ] Update docs
- [ ] Ensure all `listen()` instances behave the same...

### DIFF
--- a/modules/createHistory.js
+++ b/modules/createHistory.js
@@ -65,10 +65,10 @@ function createHistory(options={}) {
     })
   }
 
-  function listen(listener) {
+  function listen(listener, callImmediate = true) {
     changeListeners.push(listener)
 
-    if (location) {
+    if (location && callImmediate) {
       listener(location)
     } else {
       let location = getCurrentLocation()


gaearon:
This doesn't feel right to me. I think we shouldn't complicate one of the most important methods in the library by introducing a parameter that does something the consumer can always do. If it's not elegant you can always use a helper like [`_.after`](https://lodash.com/docs#after).

jamischarles:
Alright. This was mainly a question of how to best accomlpish this. Thanks Dan.

PR #99, opened 2015-10-13
ignore sessionStorage exceptions

dogosaur:
To fix this: https://github.com/rackt/history/issues/97

I don't know if this is the solution you want, but thought I'd have a go anyway ;)

### DIFF
--- a/modules/DOMStateStorage.js
+++ b/modules/DOMStateStorage.js
@@ -2,7 +2,6 @@
 import warning from 'warning'
 
 const KeyPrefix = '@@History/'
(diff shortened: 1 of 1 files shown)


taion:
Do we really want these to be no-ops in production?

What about catching the error but deferring throwing it until after we're done doing everything else?

taion:
I think #104 is touching on a lot of the same code and is more robust. Can we close this as a dupe?

mjackson:
Ya, let's follow up in #104

PR #101, opened 2015-10-14
Add forceRefresh option to createBrowserHistory

songawee:
Also, was wondering if there was a contributing.md somewhere I should be looking at?


songawee:
Darn it, I think this is a dupe.

taion:
You were beaten by #100. Sorry.

PR #102, opened 2015-10-14
enforce 2 space indentation

pwmckenna:
based on @taion's [observation that 3 space indentation wasn't detected](https://github.com/rackt/history/pull/100#discussion_r41945748). might make sense to factor out the eslint config into a rackt repo, then have all repos extend it rather than updating each one. thoughts?

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -13,6 +13,9 @@
     "array-bracket-spacing": [2, "always"],
     "comma-dangle": [2, "never"],
     "eol-last": 2,
+    "indent": [2, 2, {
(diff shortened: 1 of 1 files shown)


taion:
It's actually pretty common to not indent `case` statements. I'd rather we configure the `indent` rule as such here, rather than change the existing code. See http://eslint.org/docs/rules/indent.html

pwmckenna:
makes sense, will update

mjackson:
Looks good. Thanks, @pwmckenna! :D

PR #103, opened 2015-10-14
allow histories to cancel location updates

jlongster:
This is spawned from #100. Right now if a browser doesn't support `pushState` (or the user is forcing full refreshes), the router still immediately calls listeners, meaning react-router still tries to immediately render the new location within the same page, even though a full reload has been triggered. This causes glitches when navigating: the page changes (most likely to an invalid state) and then a second later the whole thing reloads.

This allows `finishTransition` in `createBrowserHistory` to return `false` when reloading the whole page to cancel location updates, so nothing happens.

### DIFF
--- a/modules/createHistory.js
+++ b/modules/createHistory.js
@@ -114,8 +114,9 @@ function createHistory(options={}) {
         return // Transition was interrupted.
 
       if (ok) {
-        finishTransition(nextLocation)
-        updateLocation(nextLocation)
+        if(finishTransition(nextLocation) !== false) {
+          updateLocation(nextLocation)
+        }
       } else if (location && nextLocation.action === POP) {
         let prevIndex = allKeys.indexOf(location.key)
         let nextIndex = allKeys.indexOf(nextLocation.key)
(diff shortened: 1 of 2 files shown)


taion:
Could you include the update to `createBrowserHistory` that makes it use this behavior as well, when HTML5 history isn't supported?

jlongster:
Done!

taion:
LGTM. The only potential concern I have is using `false` as the sentinel, but I can't think of anything better.

I'll let someone else merge this as well :D

modules/createHistory.js
```diff
@@ -114,8 +114,9 @@ function createHistory(options={}) {
         return // Transition was interrupted.
 
       if (ok) {
-        finishTransition(nextLocation)
-        updateLocation(nextLocation)
+        if(finishTransition(nextLocation) !== false) {
```
gaearon: Nit: missing space after `if`

mjackson:
Thanks, @jlongster! Great call.

And thanks for the review, @taion and @gaearon. I'll manually rebase and merge.

mjackson:
Merged in 0148f1f5a828cb8414d1de4a9d9d0159d55f6c71

PR #104, opened 2015-10-15
Fix saveState to handle QuotaExceededError

Nicktho:
This patch:

• Fixes https://github.com/rackt/history/issues/84 by removing the oldest state in sessionStorage before attempting to saveState again.

### DIFF
--- a/modules/DOMStateStorage.js
+++ b/modules/DOMStateStorage.js
@@ -8,18 +8,33 @@ function createKey(key) {
   return KeyPrefix + key
 }
 
+function isHistoryKey(key) {
+  return key.indexOf(KeyPrefix) === 0
+}
+
 export function saveState(key, state) {
   try {
     window.sessionStorage.setItem(createKey(key), JSON.stringify(state))
   } catch (error) {
-    if (error.name === QuotaExceededError || window.sessionStorage.length === 0) {
-      // Probably in Safari "private mode" where sessionStorage quota is 0. #42
-      warning(
-        false,
-        '[history] Unable to save state; sessionStorage is not available in Safari private mode'
-      )
+    if (error.name === QuotaExceededError) {
+      if (window.sessionStorage.length === 0) {
+        // Probably in Safari "private mode" where sessionStorage quota is 0. #42
+        warning(
+          false,
+          '[history] Unable to save state; sessionStorage is not available in Safari private mode'
+        )
 
-      return
+        return
+      }
+
+      // Reached quota, remove first (oldest) item in storage
+      for (const storedKey in window.sessionStorage) {
+        if (isHistoryKey(storedKey)) {
+          window.sessionStorage.removeItem(storedKey)
+
+          return saveState(key, state)
+        }
+      }
     }
 
     throw error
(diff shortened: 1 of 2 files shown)


Nicktho:
I went with patching `window.sessionStorage.setItem` to test this instead of overloading it and manually causing the error. Open to suggestions on other ways to test it :)

taion:
That's a cool approach. We should be using something like the MDN check here: https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API/Using_the_Web_Storage_API#Testing_for_support_vs_availability to detect for when `sessionStorage` isn't available at all, though, rather than just checking `length`.

modules/DOMStateStorage.js
```diff
@@ -8,18 +8,33 @@ function createKey(key) {
...
+        warning(
+          false,
+          '[history] Unable to save state; sessionStorage is not available in Safari private mode'
+        )
+
+        return
+      }
+
+      // Reached quota, remove first (oldest) item in storage
+      for (const storedKey in window.sessionStorage) {
```
mjackson: This loop does not necessarily start with the oldest entry first. AFAICT (just by testing manually in various browser consoles) the iteration order here is not defined. Chrome appears to iterate keys in alphabetical order, while Safari uses a random order.

Since we can't rely on browsers for consistent behavior here, one idea I had was that we could keep our own array in memory, ordered however we want (probably by least recently used) and then just try to persist that array to `sessionStorage` each time. This may be more overhead because we would be serializing a bunch of the same data every time, but it would only be as slow as `JSON.stringify` is on a potentially large array (probably not noticeable).

Another option would be to somehow specify a timestamp in the key itself, so that when we need to remove one we just grab all the keys and sort them to tell which one is the oldest.

Nicktho: > Since we can't rely on browsers for consistent behavior here, one idea I had was that we could keep our own array in memory, ordered however we want (probably by least recently used) and then just try to persist that array to sessionStorage each time...

Originally I was going to go with this idea. Before I do, there's also this method here on the storage object: https://developer.mozilla.org/en-US/docs/Web/API/Storage/key which from my testing seems to always return the correct key - ie `.key(0)` will return the oldest key in storage.

mjackson: It seems like `sessionStorage.key` isn't reliable either, at least not in Chrome.

screen shot 2015-10-15 at 6 08 40 pm

Nicktho:
Ah it looks like it is actually sorted. I'll go with the array method.

mjackson:
:+1: Thanks, @Nicktho!

Nicktho:
Sorry for the hold up guys, will be picking this up over this weekend.

mjackson:
No problem, @Nicktho. Any updates?

masonicboom:
Note that any access to window.sessionStorage (including `window.sessionStorage.length === 0`) will throw an error in iOS Safari when cookies are disabled (see my PR at https://github.com/rackt/history/pull/120).

taion:
@Nicktho Any updates here?

taion:
@Nicktho I'm going to close this for now, because I think the concept of a session-persistent array of location keys in https://github.com/rackt/history/issues/157 will give us a good way forward to implement this.

PR #108, opened 2015-10-23
Handle trailing slashes in basename

davis:
currently:
```js
// this works
history = useBasename(createHistory)({
  basename: '/base/url'
})
```

```js
// but this doesn't
history = useBasename(createHistory)({
  basename: '/base/url/'
})
```

this pr handles both cases

### DIFF
--- a/modules/useBasename.js
+++ b/modules/useBasename.js
@@ -29,7 +29,10 @@ function useBasename(createHistory) {
       if (typeof path === 'string')
         path = parsePath(path)
 
-      const pathname = basename + path.pathname
+      const pname = path.pathname
+      const normalizedBasename = basename.slice(-1) === '/' ? basename : basename + '/'
+      const normalizedPathname = pname.charAt(0) === '/' ? pname.slice(1) : pname
+      const pathname = normalizedBasename + normalizedPathname
 
       return {
         ...path,


taion:
Do we really need this in the code though? I feel like it'd be easier just to tell people to not put a trailing slash in `basename`.

davis:
Intuitively I feel like if the basenames: `'/'` and `''` have the same behavior than so should `/base` and `/base/`

mjackson:
I'd be ok with this change. There are failing tests that need to be fixed, though. Can you please fix them first, @davis?

taion:
Test failure is not related to the PR I don't think - I'm re-running the build.

taion:
Oh, never mind, I didn't scroll down far enough - those are real test failures.

davis:
Hey guys, I've changed the code to use slice/charAt instead of startsWith/endsWith so that the tests pass in all environments.

taion:
Thanks! Do we need to update any docs or changelogs to go with this?

mjackson:
We can merge this for now, @taion. I'll add a "contributing" doc that explains how to update `CHANGES.md` when making PRs.

mjackson:
Thanks @davis !

PR #109, opened 2015-10-23
Look for protocol/host at the start of paths only.

rryan:
extractPath erroneously detected a fully-qualified URL when a URL-like string
was in the path (i.e. in the query).


taion:
Nice catch. Thanks!

PR #110, opened 2015-10-23
sessionStorage security error

geirlandro:
Our app runs in an iframe and when third party cookies are turned off, we get a security error.
Added try catch around code using sessionStorage as a quick fix.


geirlandro:
Oops... Sorry, accidentally opened a pull request

Issue #113, opened 2015-10-27
Url as first parameter?


th0r:
Sorry if this question has already been raised but didn't find it.

Is there any reason for `state` parameter to be the first argument in `pushState` and `replaceState` methods?
IMHO in the most cases it will be `null`, so maybe make `path` parameter the first one?

knowbody:
@th0r I think the reason for it is the original browser history implementation, see https://developer.mozilla.org/en-US/docs/Web/API/History_API#The_pushState()_method

th0r:
@knowbody Yes, I thought about it too, but there is no `title` parameter in the `history`'s API, so it not completely reflects the original one.
And here the question arises: do we really want to reflect it?

th0r:
One can, of course, write his own middleware to swap them, but IMHO it could be the part of the core.

knowbody:
yeah I think it would be good to have a path as a first argument, I agree on that. In most cases I use `pushState(null, '/mypath')`. Was just trying to explain where it came from :)

mjackson:
@knowbody is right. `state` is the first argument because it's also the first arg to native `pushState`. We ignore the `title` arg because all browsers ignore that arg as well.

I don't think it makes any sense to have `path` as the first arg to a method called `pushState`, but we could possibly introduce `push` and `replace` methods that only take a `path` and nothing else. Implementation would be super simple:

```js
function push(path) {
  pushState(null, path)
}
```

knowbody:
> we could possibly introduce push and replace methods that only take a path and nothing else.

:point_up:  I like that

th0r:
Then +1 for `push` and `replace`

th0r:
For those, who really miss these methods, here is the middleware to add them:
```js
/**
 * Adds `history.push()` and `history.replace()` methods that allows not to specify `state` object:
 *
 * `history.pushState(null, '/profile')` => `history.push('/profile')`
 *
 * @param {function} createHistory
 * @returns {object} - `history` object
 */
function usePushReplace(createHistory) {
    return opts => {
        const history = createHistory(opts);

        /**
         * Adds new history entry
         *
         * @param {string} path
         * @param {object} [state = null]
         */
        function push(path, state = null) {
            return history.pushState(state, path);
        }

        /**
         * Replaces current history entry
         *
         * @param {string} path
         * @param {object} [state = null]
         */
        function replace(path, state = null) {
            return history.replaceState(state, path);
        }

        return {
            ...history,
            push,
            replace
        };
[...]

PR #118, opened 2015-10-29
Adds `history.push()` and `history.replace()` methods

th0r:
This allows to omit `state` parameter: `history.pushState(null, '/home') => history.push('/home')`

See #113 for more info.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
+node_modules
 lib
 umd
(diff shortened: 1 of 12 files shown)


knowbody:
@th0r looks good to me! could you just remove the `.editorconfig` we don't need it. and also please squash your commits.

Thanks a lot for that!

th0r:
Done

knowbody:
thank you!

th0r:
@knowbody No problem!)

PR #120, opened 2015-10-29
Fail gracefully when security settings prevent access to sessionStorage

masonicboom:
This can happen (at least) in iOS 9 Mobile Safari, when "Block Cookies" is set to "Always Block" (https://github.com/angular-translate/angular-translate/issues/629 suggests this has been true since iOS 7.1.2).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,7 @@
+## HEAD
+
+- Fail gracefully when Safari security settings prevent access to window.sessionStorage
+
 ## [v1.13.0]
 > Oct 28, 2015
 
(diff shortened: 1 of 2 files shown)


taion:
This is the same as #99. I think we just want to handle everything in #104.

mjackson:
This is a different error, @taion. #104 is about handling errors when the quota is actually exceeded. This PR is about handling errors when cookies are disabled in a specific environment.

taion:
It's not the same as #104, but it's a more targeted version of the same check as #99, geared toward the same thing. I guess we'd merge these separately and resolve the conflicts later?

mjackson:
Thanks, @masonicboom!

@taion Yes, we'll resolve and merge #104 separately.

PR #124, opened 2015-11-02
Use dev-expression Babel plugin

taion:
This drops the size of `History.min.js` from 27,016 B to 25,851 B.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
   },
   "bugs": "https://github.com/rackt/history/issues",
   "scripts": {
-    "build": "babel ./modules --stage 0 --loose all -d lib --ignore '__tests__'",
+    "build": "babel ./modules --stage 0 --loose all --plugins dev-expression -d lib --ignore '__tests__'",
     "build-umd": "NODE_ENV=production webpack modules/index.js umd/History.js",
     "build-min": "NODE_ENV=production webpack -p modules/index.js umd/History.min.js",
     "lint": "eslint modules",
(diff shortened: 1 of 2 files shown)


mjackson:
So we can already take advantage of this plugin without any code modifications? We're not currently using `__DEV__` anywhere. What code is it removing?

taion:
It's these two: https://github.com/4Catalyzer/babel-plugin-dev-expression#invariant, https://github.com/4Catalyzer/babel-plugin-dev-expression#warning

For a production build, this strips out calls to `warning` and removes the error messages for `invariant`.

Though https://github.com/rackt/history/blob/v1.13.0/package.json#L13 means that these will also happen for our un-minified UMD bundle, which I'm not sure if you want.

taion:
And I have absolutely no idea why this isn't broadly used in the React ecosystem outside of Facebook's own modules. It's a free win for package size if you're using `warning` and `invariant`. You'd think the people who pulled out `warning` and `invariant` from `fbjs` would have noticed that there was this great Babel plugin that goes with them.

taurose:
:+1: Thanks for this @taion . Finally a real solution for  https://github.com/rackt/react-router/pull/556

mjackson:
cc @BerkeleyTrue who maintains the `warning` package

taion:
I think this is the last missing piece of the `warning`/`invariant` trinity - the Babel plugin that Facebook use to strip everything out for production builds. Not a massive space savings by any means, but it's "free", more or less.

BerkeleyTrue:
Yes?

PR #132, opened 2015-11-10
Add back housekeeping updates

taion:
https://github.com/rackt/react-router/pull/2517

This adds back the coverage reporting for the time being.


mjackson:
Please narrow down this PR to only include the coverage fixes.

PR #137, opened 2015-11-12
Use whitelist instead of blacklist for packaging

taion:


mjackson:
What's the upside here? Just clearer?

taion:
It's more future-proof and less error-prone.

With time, the random `.whatever` files always end up piling up, and people _never_ forget to update `.npmignore` to drop them.

The setup for `history` is actually fine right now (except `.npmignore` is shipped for some reason), but the one for React Router visibly hit bit rot before I made the corresponding change there. If you take a look at the tarball for `react-router@1.0.0-beta4`, it includes among other things the `coverage` directory (isparta output), `.eslintrc`, `.babelrc`, and `.travis.yml`.

The idea is that if you explicitly whitelist things, you can't possibly mess up and end up with cruft in your package.

taion:
Case in point, this is actually already semi-misconfigured due to https://github.com/rackt/history/pull/136, because I forgot to update `.npmignore` in that PR after splitting it out from this one.

In practice it's probably not a problem unless you've run the tests locally with `COVERAGE=true`, but you would have ended up with `coverage/` in the npm tarball if you'd published after doing so.

I've just found that in practice, I end up with random stuff that I don't want to ship to npm much, much more often than I end up with new directories that I do want to ship.

taion:
Merging this to sync config with React Router.

PR #140, opened 2015-11-13
Run tests in IE 9

taion:


modules/__tests__/HashHistory-test.js
```diff
@@ -26,11 +26,18 @@ describe('hash history', function () {
...
-  if (supportsGoWithoutReloadUsingHash()) {
+  if (supportsHistory()) {
+    describePopState(createHashHistory)
+  } else {
+    describe.skip(null, function () {
+      describePopState(createHashHistory)
+    })
+  }
+
+  if (supportsHistory() && supportsGoWithoutReloadUsingHash()) {
```
taion: I'm -0 on doing the check here instead of in `supportsGoWithoutReloadUsingHash`. It could go in the latter, but I don't see much value in warning all IE9 users about this all the time, since it's not like there's a choice to use anything other than hash history there.

mjackson:
It's been a while since I tried to dig in here and figure out what exactly was failing on IE9. It looks like the only tests that were failing were in `describePopState`. Is that correct, @taion?

In any case, thank you for digging in here! It's been on my TODO list for months.

taion:
It's `describePopState` and the two tests under `supportsGoWithoutReloadUsingHash()` - they all need to be skipped to avoid a full reload on IE9.

mjackson:
Awesome. I'm :100: on this. Thanks again!

taion:
No problem.

Do you agree with adding that branch in the test, rather than also checking for HTML5 History support in `supportsGoWithoutReloadUsingHash`?

mjackson:
Ya, I think so. That method has such a long name because I was trying to be as descriptive as possible. I'd prefer it to only do that one thing and nothing else.

PR #146, opened 2015-11-19
Disable browser history on Chrome iOS

timdorr:
Fixes rackt/react-router#2565


timdorr:
Unfortunately, automated testing with Chrome on iOS is impossible. But in my tests on my iPad under iOS 9, this works.

taion:
LGTM.

Ideally if this is a CriOS bug, we can take this out at some point once it's fixed there. It's not _so_ broken that this is a dealbreaker, but this is an improvement over the status quo.

taion:
Shoot, one thing - could you add an entry to `CHANGES.md`?

PR #147, opened 2015-11-20
Add FIXME note to CriOS workaround

taion:
@timdorr Do you mind if I make this change? Makes it easier to spot this as potentially our bug in the code, given that other HTML5 history examples work on Chrome iOS.


taion:
Gonna merge this for now just to clarify the note.

PR #152, opened 2015-11-20
Add ES2015 module build

taion:


taion:
Sync w/https://github.com/rackt/react-router/pull/2530.

PR #153, opened 2015-11-22
Support anchors in the hash history

just-boris:
Now `location.hash` is ignored when `createHashHistory` is used.

This fix makes the behavior of `createHashHistory` more compatible with `createBrowserHistory`.

### DIFF
--- a/modules/DOMUtils.js
+++ b/modules/DOMUtils.js
@@ -17,7 +17,8 @@ export function removeEventListener(node, event, listener) {
 export function getHashPath() {
   // We can't use window.location.hash here because it's not
   // consistent across browsers - Firefox will pre-decode it!
-  return window.location.href.split('#')[1] || ''
+  let href = window.location.href
+  return href.indexOf('#') > -1 ? href.substring(href.indexOf('#') + 1) : ''
 }
 
 export function replaceHashPath(path) {
--- a/modules/__tests__/HashHistory-test.js
+++ b/modules/__tests__/HashHistory-test.js
@@ -9,6 +9,7 @@ import describeReplaceState from './describeReplaceState'
 import describeReplace from './describeReplace'
 import describePopState from './describePopState'
 import describeQueryKey from './describeQueryKey'
+import describeHashSupport from './describeHashSupport'
 import describeBasename from './describeBasename'
 import describeQueries from './describeQueries'
 import describeGo from './describeGo'
@@ -25,6 +26,7 @@ describe('hash history', function () {
   describePush(createHashHistory)
   describeReplaceState(createHashHistory)
   describeReplace(createHashHistory)
+  describeHashSupport(createHashHistory)
   describeBasename(createHashHistory)
   describeQueries(createHashHistory)
 
@@ -49,5 +51,22 @@ describe('hash history', function () {
   it('knows how to make hrefs', function () {
     let history = createHashHistory()
     expect(history.createHref('/a/path')).toEqual('#/a/path')
+  })
+
+  it('serializes hash from location', function () {
+    let history = createHashHistory({ queryKey: false })
+    history.pushState(null, '/home?the=query#anchor')
+    expect(window.location.hash).toEqual('#/home?the=query#anchor')
+  })
+
+  it('reads hash from location', function (done) {
+    window.location.hash = '#/home#anchor'
+    let history = createHashHistory()
+    let unlisten = history.listen(function (location) {
+      expect(location.pathname).toEqual('/home')
+      expect(location.hash).toEqual('#anchor')
+      done()
+    })
+    unlisten()
   })
 })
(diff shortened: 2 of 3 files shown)


taion:
Why? They're not going to do anything. You also didn't update `getCurrentLocation` either.

just-boris:
Thanks for noting this, I fixed `getCurrentLocation` too.

I am going to use it for anchorScroll behavior. I implemented this in my app, and now it works nice, except that hash does not appear in location, so I can't share a link for someone else.

modules/DOMUtils.js
```diff
@@ -17,7 +17,7 @@ export function removeEventListener(node, event, listener) {
 export function getHashPath() {
   // We can't use window.location.hash here because it's not
   // consistent across browsers - Firefox will pre-decode it!
-  return window.location.href.split('#')[1] || ''
+  return window.location.href.split('#').slice(1).join('#') || ''
```
taion: Use `indexOf` and `substr` to avoid extra work with arrays. It doesn't lose at all on readability.

taion:
Thanks. One more nit. I'm going to hold off on merging - code looks fine but gonna see what others think about adding this feature.

just-boris:
Done. 
Okay, I will be wait

taion:
Per http://stackoverflow.com/questions/10850781/multiple-hash-signs-in-url, `#` is not actually legal in a URL hash. Interesting - do you think it'd be okay to escape it, or would it make your URLs unnecessarily ugly?

just-boris:
Ok, that makes sense. I checked that other routing library, like `Backbone.router` does this too.
I am going to keep updated `getHashPath()` function to support manually typed URLs, also I will add escaping into `createHref ` and unescaping in the `getCurrentLocation `

I don't have time to do it now but I'll do it during next week.

taion:
I'm struggling to think of a good way to do this, though - we don't do any URI encoding/decoding elsewhere in this library. Maybe this is just okay as-is, since it seems to work properly on all the browsers we care about anyway.

just-boris:
Yes, that should work fine as it is now in all supported browsers. If something will go wrong, tests will show it.
So, should I update my PR?

taion:
What would there to be update if so?

just-boris:
If we are going to escape, it should be added in the `createHref ` to get escaped links, and `getCurrentLocation` to parse them back.

It will be something like this
```diff
function createHref(path) {
+    // '#' is not a valid symbol in the hash according to RFC-3986    
+    path = path.replace('#', '%23')
      return '#' + history.createHref(path)
}
```

taion:
I'm not sure we should escape. We're not otherwise escaping anything else in the hash string, so I'm concerned about inconsistency.

just-boris:
Ok, so I leave this as it is now.
Is there something else blocking merge of this?

taion:
I want @mjackson to take a look.

taion:
Ping @mjackson.

mjackson:
I'm not convinced this is necessary. Every day hash history becomes less and less relevant. Also, the double hash results in an invalid URL.

taion:
@mjackson I think for public-facing sites, using hash history these days is an anti-pattern. However, it's really convenient to have hash history for setting up one-shot things for e.g. internal use. I use this in a couple of places for internal-facing tools where my front end is purely just a bunch of static files, i.e. when I just want a super-minimal hacky front end to some API I'm building out or something.

just-boris:
@mjackson I am agree with @taion, I am using hash history on my project on Github Pages. I can't change server-side logic there. So  I am using my fork to do client-only routing

taion:
@just-boris For GH pages, the preferred solution is to render everything statically, then upload the static results to GH. This is what we do for e.g. React-Bootstrap. It's not too bad.

just-boris:
@taion I am not talking about a static site with React. I have a SPA that renders result from Github API. Here is it: http://catatron.com/revive/
Now I don't see any solution for me rather than this PR.

taion:
Oh that is quite cool. I think this broadly makes sense to me – though note that we don't currently have scroll behavior management for anchors yet anyway.

just-boris:
@taion yes, I made it up on my side. It doesn't require to change anything in the core library.
Source code is [here](https://github.com/just-boris/revive/blob/master/src/application/util/useAnchorScroll.js) I can share it if someone will be interested, but now we have to finish this first.

SimenB:
Any news here?

mjackson:
I believe this should work on master, as of f12d632691ebef652d0cf5dd67a033e87746fc4a. We'll release a v3 beta soon (hopefully today) that you can test.

just-boris:
All right, the current master should work as expected.

So, closing this PR and waiting for  3.0 release then. According to changelog, it seems that it will not be difficult to upgrade

PR #154, opened 2015-11-22
Fix grammar in QuerySupport.md

yang-wei:


taion:
This part of the docs is actually outdated - `push` and `replace` only support a single argument. I'm going to update separately.

yang-wei:
@taion That's why `push` dint work for me as in documentation. I can write the documentation if you don't mind.

taion:
https://github.com/rackt/history/pull/155

PR #156, opened 2015-11-23
Configure the coverage reporter

taion:


taion:
"Why isn't Coveralls updating?" "D'oh!"

PR #160, opened 2015-11-29
Fixed bug in code example

coryhouse:
Old example was returning a string instead of a bool. Now prompting for confirmation via confirm, which I assume was the intent of the example.


taion:
No, just like on a before unload hook, you're supposed to return a string to get the default confirm box.

coryhouse:
Ah, sorry, was about to close this but you beat me to it. :)

PR #167, opened 2015-12-01
Consider the full path in changing PUSH to REPLACE

### LINKED ISSUE #166: Same URL with different hash anchor tags are not maintained in history properly
I have a case where I have a bunch of anchors on a page, and want to record when a user navigates between those anchors. With the change in PR #43, it removes that functionality, because now it treats all hash changes on the same page as if they're one single item in the browser history. 

I believe this might be a regression bug according to @taion.

taion:
Fixes #166

Organizing the changelog a bit ahead of adding something for #141 to hopefully make it a bit more usable.

### DIFF
--- a/modules/createHistory.js
+++ b/modules/createHistory.js
@@ -117,11 +117,10 @@ function createHistory(options={}) {
       if (ok) {
         // treat PUSH to current path like REPLACE to be consistent with browsers
         if (nextLocation.action === PUSH) {
-          let { pathname, search } = getCurrentLocation()
-          let currentPath = pathname + search
-          let path = nextLocation.pathname + nextLocation.search
+          const prevPath = createPath(location)
+          const nextPath = createPath(nextLocation)
 
-          if (currentPath === path)
+          if (nextPath === prevPath)
             nextLocation.action = REPLACE
         }
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,11 +1,16 @@
 ## [HEAD]
 
+#### Bug fixes
 - Disable browser history on Chrome iOS ([#146])
+- Do not convert same-path PUSH to REPLACE if the hash has changed ([#167]) 
+
+#### Other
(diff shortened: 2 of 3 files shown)


modules/createHistory.js
```diff
@@ -117,11 +117,10 @@ function createHistory(options={}) {
       if (ok) {
         // treat PUSH to current path like REPLACE to be consistent with browsers
         if (nextLocation.action === PUSH) {
-          let { pathname, search } = getCurrentLocation()
-          let currentPath = pathname + search
-          let path = nextLocation.pathname + nextLocation.search
+          const prevPath = createPath(location)
```
taion: This is not 100% correct, since it doesn't e.g. account for `basename` when using browser history. In practice that's not an issue, though, since `basename` doesn't change.

This should be the "final" `createPath`, but I can't access that here.

taurose:
LGTM. Thanks for tackling that.

As for your question why I used `getCurrentLocation()`, I think I wanted to avoid race conditions and mimic the default browser behavior as close as possible. Right now I can't think of any realistic condition where using `location` instead would be a problem, so I'm good with that change.

taion:
I believe in a base `history`, transitions are fully synchronous, so I don't think race conditions are an issue. Let's go with this for now and patch later if needed.

taion:
@taurose I think there might be a separate related race condition when using async routes in `ReactRouter.useRoutes`, though - i.e. where the transition is async, so the user can e.g. change their mind and try to `PUSH` a separate location while the old routing state is still shown... for browser-mimicking that should get changed to a `REPLACE`, but we can't even handle that in `history` since the route state management is all downstream.

taurose:
@taion Yeah that seems like a feature we lost from 0.13. Since history doesn't deal with data loading (or anything that happens after the URL changes), I'd say this should be handled by the router.

taion:
Yeah, the router can just use `listenBefore`.

PR #171, opened 2015-12-06
Use query-string instead of qs

### LINKED ISSUE #121: Use query-string instead of qs
Just noticed that history is 60k before minification!

screen shot 2015-10-29 at 6 12 49 pm

mjackson:
Fixes #121

### DIFF
--- a/modules/useQueries.js
+++ b/modules/useQueries.js
@@ -1,15 +1,25 @@
-import qs from 'qs'
+import warning from 'warning'
+import { parse, stringify } from 'query-string'
 import runTransitionHook from './runTransitionHook'
 import parsePath from './parsePath'
 
 const SEARCH_BASE_KEY = '$searchBase'
 
 function defaultStringifyQuery(query) {
-  return qs.stringify(query, { arrayFormat: 'brackets' }).replace(/%20/g, '+')
+  return stringify(query).replace(/%20/g, '+')
 }
 
-function defaultParseQueryString(queryString) {
-  return qs.parse(queryString.replace(/\+/g, '%20'))
+const defaultParseQueryString = parse
+
+function isNestedObject(object) {
+  for (const p in object)
+    if (object.hasOwnProperty(p) &&
+        typeof object[p] === 'object' &&
+        !Array.isArray(object[p]) &&
+        object[p] !== null)
+      return true
+
+  return false
 }
 
 /**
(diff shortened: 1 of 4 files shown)


modules/useQueries.js
```diff
@@ -1,15 +1,22 @@
...
+  return stringify(query).replace(/%20/g, '+')
 }
 
-function defaultParseQueryString(queryString) {
-  return qs.parse(queryString.replace(/\+/g, '%20'))
+const defaultParseQueryString = parse
+
+function isNestedObject(object) {
+  for (var p in object)
+    if (typeof object[p] === 'object')
```
taion: How do you want to handle arrays and `null` here? query-string handles them just fine (though stringified differently from qs in both cases).

modules/useQueries.js
```diff
@@ -1,15 +1,22 @@
...
-  return qs.stringify(query, { arrayFormat: 'brackets' }).replace(/%20/g, '+')
+  return stringify(query).replace(/%20/g, '+')
 }
 
-function defaultParseQueryString(queryString) {
-  return qs.parse(queryString.replace(/\+/g, '%20'))
+const defaultParseQueryString = parse
+
+function isNestedObject(object) {
+  for (var p in object)
```
taion: Nits:
- This should probably be `const`
- `hasOwnProperty` guard? (ESLint can actually be set to warn against that)

modules/useQueries.js
timdorr: It looks like it handles those sanely. This is just a developer convenience warning, so they're aware we're not looking to handle nested queries. I wouldn't try to go overboard on supporting a bunch of edge cases here. I think they'll figure it out when they see `%5Bobject+Object%5D` in their URLs :smile:

timdorr:
@mjackson Feel free to smash that commit. I don't know how you like your commit history. I'm more of a multiple-commit kind of dude, but you may want it clean.

taion:
@timdorr Regarding your line comment that got eaten, what I meant was we should probably _not_ warn in those cases, because `query-string` is sane about handling them.

timdorr:
Duh, because arrays are typeof object. I always forget that. Let me amend that commit then.

taion:
Arrays _and_ `null`. I usually just do `p instanceof Object`.

timdorr:
But `[] instanceof Object === true`

taion:
D'oh. Let's squash and merge this then.

PR #173, opened 2015-12-06
Location descriptors in createPath and createHref

taion:


taion:
Codemod: https://github.com/rackt/rackt-codemod/blob/v0.1.0/modules/history/deprecate-createPath-createHref-query.js

timdorr:
@taion Your warnings are backwards. The first arg to `warning` should be truthy if you want it to not show the warning. The query should be *falsey/undefined* when they are *not* using the deprecated form.

taion:
Oops! Let's make a mental note to fix those when we add them back.

PR #177, opened 2015-12-07
Reduce saveState data in sessionStorage

koba04:
I'm using history with react-router. Thank you for your great work!

I'm using Session Storage wth my app.
So I often confirm the developer console for debugging app data in Session Storage.

history is adding a state to sessionStorage every history changes.
As a result, my Session Storage is filled by history's state regardless of using it.
This behavior makes my debugging hard.

screen shot 2015-12-07 at 14 01 21

My PR is for that it saves history's state in Session Storage only when history's state is not empty.

Thanks!

### DIFF
--- a/modules/DOMStateStorage.js
+++ b/modules/DOMStateStorage.js
@@ -11,7 +11,11 @@ function createKey(key) {
 
 export function saveState(key, state) {
   try {
-    window.sessionStorage.setItem(createKey(key), JSON.stringify(state))
+    if (state == null) {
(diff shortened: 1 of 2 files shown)


taion:
Can you add some tests around this?

koba04:
@taion 
OK, I'll work on it!

koba04:
@taion 
I've added some tests for DOMStateStorage.
Is this not enough?

taion:
Thanks!

koba04:
Thanks!

PR #179, opened 2015-12-07
Use forced REPLACE only when both location and state are the same

Macroz:
Fixes https://github.com/rackt/history/issues/178

### DIFF
--- a/modules/createHistory.js
+++ b/modules/createHistory.js
@@ -121,7 +121,7 @@ function createHistory(options={}) {
           const prevPath = createPath(location)
           const nextPath = createPath(nextLocation)
 
-          if (nextPath === prevPath)
+          if (nextPath === prevPath && deepEqual(location.state, nextLocation.state))
             nextLocation.action = REPLACE
         }
 
--- a/modules/__tests__/describePush.js
+++ b/modules/__tests__/describePush.js
@@ -116,6 +116,43 @@ function describePush(createHistory) {
             history.push({
               pathname: '/home',
               search: '?the=query',
+              state: { the: 'state' }
+            })
+          },
+          function (location) {
+            expect(location.pathname).toEqual('/home')
+            expect(location.search).toEqual('?the=query')
+            expect(location.state).toEqual({ the: 'state' })
+            expect(location.action).toEqual(REPLACE)
+          }
+        ]
+
+        unlisten = history.listen(execSteps(steps, done))
+      })
+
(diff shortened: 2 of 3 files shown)


taion:
Please add test coverage here.

Macroz:
Yes, will add, if this is considered a useful change.

modules/createHistory.js
```diff
@@ -120,7 +120,7 @@ function createHistory(options={}) {
           const prevPath = createPath(location)
           const nextPath = createPath(nextLocation)
 
-          if (nextPath === prevPath)
+          if (nextPath === prevPath && deepEqual(location.state, nextLocation.state))
```
taion: We should probably make this check w/something that shares code w/`locationsAreEqual`, possibly prior to calling `createPath`.

package.json
```diff
@@ -7,6 +7,7 @@
     "docs",
     "es6",
     "lib",
+    "modules",
```
mjackson: Please omit this line.

mjackson:
Logically, it seems like this PR is on the right track and philosophically compatible with the fix that @taurose made in #43. The idea behind #43 is that when the exact same `location` is pushed, we should use replace instead of push. But since `state` is part of a `location` object (i.e. `location.state`) we should also consider whether the state is the same when making that decision.

@Macroz Any reason you chose to modify package.json? Otherwise, I'd say this looks good.

taion:
@mjackson 

One question I have is where this "replace PUSH with REPLACE" logic should go.

I think `transitionTo` might be the wrong place. Better would be either `push` or maybe even upstream in `<Link>` in React Router.

There's nothing in the HTML5 history API that strictly prevents users from re-pushing the exact same location. The "no extra push" behavior only comes from following `<a>`s.

Do you think a user might ever legitimately need to push the same location? Or should we just not bother supporting that?

mjackson:
> Do you think a user might ever legitimately need to push the same location?

No, I don't think that's a legit use case. You shouldn't have the ability to PUSH to where you already are.

taion:
Fair enough.

Macroz:
@mjackson I added the modules directory to use the repo as a npm package directly from GitHub, while waiting for the discussion to finish. Now I've removed it, rebased to master and squashed the commits. Does this look good?

I noticed that there are some tests failing in master at the moment.

Macroz:
Great! Thanks for the fix!

PR #186, opened 2015-12-09
Silence warnings for deprecations since 1.13

### LINKED ISSUE #185: [meta] Ship v1.16
I fixed `createLocation` in e1e0882389d47cea5f563871bee7400cd1953e59 and added the deprecation warnings for the `state` arg.

At this point I'd like to ship a version 1.16 that doesn't emit all the warnings. Then we can ship a router version 1.0.3 that uses it as a peer dep and stop breaking people's builds. I think the easiest way to do this is to just comment out all the warnings in history. This will at least get us back to a place where people who used `history: "^1.15.0"` in their package.json have a build that works.

mjackson:
This PR silences all warnings that were introduced since 1.13 in preparation for a 1.16 release.

Fixes #185

### DIFF
--- a/modules/createHistory.js
+++ b/modules/createHistory.js
@@ -1,4 +1,4 @@
-import warning from 'warning'
+//import warning from 'warning'
 import deepEqual from 'deep-equal'
 import { loopAsync } from './AsyncUtils'
 import { PUSH, REPLACE, POP } from './Actions'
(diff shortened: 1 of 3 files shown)


mjackson:
Actually, we should probably just merge into a special 1.16 branch and release from there.

taion:
:+1:

PR #189, opened 2015-12-14
Reverse createHref and createPath warnings

timdorr:
These were backwards. We should push a 1.16.1 so I can get rid of all warnings upstream in `react-router`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,9 +2,11 @@
 > Unreleased
 
 - **Bugfix:** Don't throw in memory history when out of history entries ([#170])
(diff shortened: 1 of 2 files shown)


taion:
Why not just drop them entirely for now? `master` is 1.16.0, no?

timdorr:
I'd rather not have to remember. We can just remove them as another PR and do a quick revert PR to put them back. Less error-prone that way.

taion:
Sure - can you update the changelog then?

taion:
:+1: squash and merge

PR #190, opened 2015-12-14
Silence warnings on createPath and createHref

timdorr:
OK, **this** should be in `1.16.1`. Then we can revert this PR at a later date when we want all the warnings back.


taion:
:shipit:

mjackson:
The approach I took in #186 was to just comment stuff out. I know that seems a little weird, but it means I don't have to go hunting for SHAs later when I want to put that stuff back in.

I'm gonna go ahead and push another commit that follows that pattern and ship 1.16.1. Thanks, @timdorr

timdorr:
Cool beans. However you'd like to do it is fine with me.

PR #197, opened 2016-01-01
Force basename when location doesn't include it

neverfox:
This treats basename as the minimally viable pathname. Combined with a change to redux-simple-router to replace history with the initial location, this will allow for / -> /base/, /foo -> /base/foo/, /base -> /base/, /base/foo -> /base/foo/. As it is now, it's possible for a redux app to run at both / and /base, contrary to expectations.

### DIFF
--- a/modules/useBasename.js
+++ b/modules/useBasename.js
@@ -27,7 +27,7 @@ function useBasename(createHistory) {
           if (location.pathname === '')
             location.pathname = '/'
         } else {
-          location.basename = ''
+          location.basename = basename
         }
       }
 


taion:
Post-https://github.com/rackt/redux-simple-router/pull/141 RSR no longer works that way, so this is no longer relevant.

neverfox:
But I'm working off of master of redux-simple-router and the behavior is still that way. Because basename will be '' in when someone enters, say, `localhost:3000`, and no attempt is made by RSR to replace history with the initial location, the address bar will stay `localhost:3000`. This is still relevant.

neverfox:
If you like, I can make a repo demonstrating.

taion:
Huh? That seems extremely odd and overly-specific. You're saying that, given an app with a basename of `/foo`, you want to redirect users from `/` to `/foo`?

taion:
That's just not right. If your `basename` is set to `foo`, then `/` does not map to:

```js
{
  pathname: '/',
  basename: '/foo'
}
```

"When the app inits" is not a meaningful concept here – if somehow you're handling things above the basename, that's outside the concern of the router or of the basename-configured history.

But in no case should it just lie to you.

neverfox:
Yes. That's exactly how, say, rootURL works in Ember. It seems highly odd to me for an app to work at two different states of the address bar. It should represent the minimally viable mount point for the app, and paths that are "deficient" should be corrected. Right now it works as nothing more than a way for the develop to say "*if the path includes basename, ignore basename" but it's much more natural to think of it as "if path *doesn't* include basename, make sure it does".

neverfox:
> That's just not right. If your basename is set to foo, then / does not map to

I didn't say it did. I said the app boots just fine, which is highly odd to me and contrary to how routers work in other frameworks with similar root or base url settings.

neverfox:
>  if somehow you're handling things above the basename, that's outside the concern of the router or of the basename-configured history.

If they are, they shouldn't be mounting the app yet, but the app is mounted. And if it is, it should utilize basename.

taion:
I suppose we could throw an exception there. I don't know that this is better than the current behavior. In no event should we report a location that's inconsistent with the URL, though.

But why are you even bothering to set a `basename` if your app is handling `/`?

neverfox:
Fair point about that not being inconsistent. It's handling because that's the out of the box behavior, which I was hoping to change by making the basename setting more of a way of declaring where an app should redirect to if something else is attempted (again similar to what other framework routers such as Ember's do). But, you're right that that's not history's job.

taion:
Yeah just redirecting to the `basename` is the right way to handle this if it comes up, though it's a bit of an odd scenario anyway.

neverfox:
In any case, thanks for taking the time to walk me through your thinking.

taion:
No problem.

Noitidart:
I think I'm having this issue. My basename is `about:mytool` and I want pages to be `about:mytool?page/1`, `about:mytool?page/2`, etc. `(new URL('about:mytool')).host` shows that there is no basename. @taion how did you handle this?

In another instance when I am running locally, my basepath should be `file:///C:/Users/Mercurius/Documents/GitHub/Non-JSX-React-Router-Redux-Demo/app.html` but router is making links go to `file:///?page1`

PR #200, opened 2016-01-06
Handle location descriptors in createLocation

taion:


modules/__tests__/describeQueries.js
```diff
@@ -201,6 +201,37 @@ function describeQueries(createHistory) {
...
+      describe('in createLocation', function () {
+        it('works with string', function () {
+          const location = history.createLocation('/the/path?the=query')
+
+          expect(location.pathname).toEqual('/the/path')
+          expect(location.query).toEqual({ the: 'query' })
+          expect(location.search).toEqual('?the=query')
+        })
+
+        it('works with object with query', function () {
```
taion: This case will not work without this patch.

taion:
@mjackson Can you look at this when you get a chance, and maybe cut 2.0.0-rc2 off this? It's blocking a few more API cleanups we need for the router.

taion:
On second thought, this might be the wrong implementation; instead, we're probably better off just passing fields through `createLocation`.

taion:
Nope, never mind, that doesn't work.

taion:
Updated for better forward-compat with https://github.com/rackt/react-router/issues/2859.

timdorr:
Looks like the changes merge conflict. That's why I tend to leave docs out of PRs. For now, just rebase and get it green :smile:

taion:
Eh, just rebased for now.

PR #202, opened 2016-01-08
F/history length

mathieudutour:
This exposes the history length and allows to reset the history (useful for an app after the onboarding or after logging out for instance).

Is it of any interest for you and I should write tests or you don't want that?


taion:
See https://github.com/rackt/react-router/pull/843, https://github.com/rackt/history/issues/89, and https://github.com/rackt/react-router/pull/1376#issuecomment-118185067.

This is not quite sufficient, so we'd rather not provide this in this form right now.

PR #203, opened 2016-01-09
postinstall → prepublish

jeremyruppel:
Currently, the `postinstall` lifecycle script fails on node 0.10.x because [child_process.execSync doesn't exist.][1] npm's [docs suggest][2] that the best practice for precompilation and transpiling is to do it in a `prepublish` script instead, which would totally avoid the issue altogether.

[1]: https://nodejs.org/docs/latest-v0.10.x/api/child_process.html
[2]: https://docs.npmjs.com/files/package.json#devdependencies


taion:
The point of this is to allow installing from GitHub. Also, we're using packaging whitelists, as is generally considered best practices these days.

jeremyruppel:
Thanks, I missed the `files` field, I'd be more than happy to roll back the .npmignore addition. But if your goal is to support installing from GitHub, doesn't checking in the generated source make more sense than relying on the client system to be able to generate it?

taion:
We will not check in the built artifacts – they're artifacts rather than source and should not be source controlled. GitHub install will remain a best-efforts thing for now.

bengl:
could you just use [`spawn-sync`](https://www.npmjs.com/package/spawn-sync) instead to make it run on node 0.10.x?

taion:
That would defeat the purpose – this is entirely just a best-efforts thing to make it possible in some cases to install from GitHub. In general you should not be installing packages that require transpilation from GitHub anyway – this isn't supported at all for e.g. React or most other packages.

jeremyruppel:
Good call @bengl. @taion how do you feel about #204?

taion:
See https://github.com/rackt/react-router/pull/2880.

I'm updating the script to move the `require` statements into the "transpiled code not found" branch.

PR #204, opened 2016-01-11
Use spawn instead of execSync

jeremyruppel:
This simply uses spawn instead of execSync to support 0.10.x versions of node.


taion:
The entire point of this is to run synchronously and fail if the spawned process fails. This completely removes that.

PR #205, opened 2016-01-13
Fix installing from GitHub

taion:


taion:
I just noticed this on the React Router end as well, but our use of `dev-expression` actually makes this even more brittle than before – but at least we have an error message now.

PR #210, opened 2016-01-16
Fix tests for new expect

taion:
cc @taurose


taurose:
Oh, hadn't noticed :smile:

PR #222, opened 2016-02-03
Fix search base logic with empty query

### LINKED ISSUE #221: Cannot submit an empty query {} object when history.push() is called.
If your `location.search = '?debug=true&tabs=Q1'` for example and you history.push(location) where `location.query = {debug: 'true'}` it will successfully remove the `tabs=Q1` part of the queryString. But if you have `location.search = '?tabs=Q1'` and you history push a location object where `query = {}` it won't change the URL at all.

I believe the issue stems from this line right here: https://github.com/rackt/history/blob/master/modules/useQueries.js#L56

taion:
Fixes #221.


timdorr:
LGTM. :+1:

mjackson:
Awesome, thanks @taion. I hope we can clean up this code when we remove the deprecations in v3.

PR #223, opened 2016-02-03
Ignore older safari QUOTA_EXCEEDED_ERRs as well

aaronjensen:
Safari 5.1 throws a different error, without this `history` enabled sites don't work at all in safari private browsing.


aaronjensen:
Actually, now that `scroll-behavior` has updated to support history 2, it seems like the site does work, it just throws the quota error. So this isn't as bad, but it'd still be nice to handle both of these. Thanks!

timdorr:
LGTM. Thanks!

PR #226, opened 2016-02-08
port loopAsync from react-router (fixes #225)

### LINKED ISSUE #225: loopAsync chokes for synchronous task + many iterations
This is kind of an edge case, stumbled on it while attaching thousands of listeners on `listenBefore`

```jsx
//this runs fine
loopAsync(100, (i, next, done) => next(i), () => console.log('done!'));

//this throws "RangeError: Maximum call stack size exceeded"
loopAsync(10000, (i, next, done) => next(i), () => console.log('done!'));
```

This seems to happen because when the handler calls `next` synchronously, [every iteration](https://github.com/rackt/history/blob/master/modules/AsyncUtils.js#L15) causes the stack to grow.  

I'm sure this is a non-issue to most, and will rarely pop up. Will try to think of a fix, unless you folks have an idea? 

Thanks for the lovely lib!

threepointone:
ref - https://github.com/rackt/history/issues/225

this fix prevents loopAsync from throwing RangeErrors when the `turns` is a high number.


taion:
Thanks!

PR #229, opened 2016-02-12
Use eslint-config-airbnb

mjackson:
This is just an experiment using AirBnB's ESLint config. I had to disable a bunch of stuff, but it's mostly just my personal preference. The big win here is we don't have to maintain our own ESLint config anymore.


.eslintrc
```diff
@@ -1,3 +1,24 @@
 {
-  "extends": "rackt"
+  "extends": "airbnb",
+  "parser": "babel-eslint",
+  "rules": {
+    "array-bracket-spacing": [ 2, "always" ],
+    "comma-dangle": [ 2, "never" ],
```
taion: This one is divisive – I used to hate it, but it does make diffs potentially be cleaner.

knowbody:
:+1: from me, I like the airbnb config

gaearon:
Why not align with React source style while we’re at it?

taion:
Isn't the Airbnb config more or less just a stricter version of the React source style? It'd be nice if there were a "React" ESLint config, though.

chicoxyzzy:
FYI airbnb's styleguide had a lot of breaking changes last couple of months and older versions aren't supported anymore. For big projects it takes a lot of time to update to newer versions.

knowbody:
@chicoxyzzy both codebases (react-router and history) are not that big though, it shouldn't take long to update them.

mjackson:
Is there an Edlinger config for react core somewhere?

mjackson:
*ESLint

knowbody:
isn't it this one: https://github.com/facebook/react/blob/master/.eslintrc

taion:
It composes in https://github.com/facebook/fbjs/blob/master/scripts/eslint/.eslintrc.js as well. Is it intentional that this isn't pulled out into a separate facebook/eslint-config-facebook or something?

I think in practice it's pretty similar to the Airbnb rule set, which is why I like it as a low-effort fallback. I have many open-source and private projects that follow eslint-config-airbnb, and have not had issues with recent changes in the past few months.

gaearon:
My only concern with using Airbnb is all those rule overrides. We end up in a lot of copy paste adjustments in different projects which is exactly what we wanted to solve with a shared config.

taion:
I'd move toward dropping those rule overrides. I run everything with either stock `airbnb`, or with only a `"eqeqeq": [2, "allow-null"]` override.

taion:
Or rather – if we e.g. don't want semicolons, we couldn't use the React code style as-is anyway.

gaearon:
It’s always the semicolons.

mjackson:
Haha. I blame @ryanflorence for my semi-colon avoidance.

Seriously tho, we could just keep our shared config on GH and not publish
to npm.

taion:
Totally selfishly, I'd prefer to have semis.

Right now I have no perceptible style shift in going between my own code and e.g. facebook/relay code, but I do feel a style shift in going to ex-Rackt code (because of semicolons and array bracket spacing).

My life would be easier if everything were consistent :laughing:

mjackson:
I'll send another PR that uses https://github.com/facebook/react/blob/master/.eslintrc and see how close we are to that.

mjackson:
Fixed in cae74e4f8cc9bfea9bd07bed304e1a9ba33f2e38

PR #230, opened 2016-02-16
Fix SCRIPT16389: Unspecified error. IE11 on page refresh

glynjonesatwork:
If there is no browser history in the window in IE11 the window.history.state throws an unspecified error. Wrap in try catch.
Similar issue found in Angular https://github.com/angular/angular.js/issues/10367

### DIFF
--- a/modules/createBrowserHistory.js
+++ b/modules/createBrowserHistory.js
@@ -28,7 +28,7 @@ function createBrowserHistory(options={}) {
   function getCurrentLocation(historyState) {
     try {
       historyState = historyState || window.history.state || {}
-    } catch {
+    } catch (e) {
       historyState = {}
     }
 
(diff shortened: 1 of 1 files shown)


ericingram:
We also have this issue. Any chance it will be fixed/merged into master?

taion:
There's a merge conflict as-is.

Can you elaborate on how to reproduce the initial issue? We do run tests against IE10 and IE11, and we're not seeing them fail.

ericingram:
The problem we're seeing is inside an iframe on IE 11+ (same domain). Adding a try catch around window.history.state solves the problem for us.

glynjonesatwork:
We found the issue when you try and refresh a page that has just been initally loaded. If the IE11 window has not navigated to any other page and has no history, this causes the problem. A particular issue for us with a single-page app.

kirillkostko:
@glynjonesatwork Could you please fix a conflict on your pull request? Thank you in advance!

taion:
BTW, note that 2.x is on the v2.x branch, not master.

taion:
Replaced by https://github.com/mjackson/history/pull/289.

PR #232, opened 2016-02-18
[WIP] Extract location creation to "location providers"

timdorr:
Hopefully, this spawns some good discussions. It comes out of these issues on router: 

reactjs/react-router#2680
reactjs/react-router#3089

I've separated out creating locations from the history state management stuff. This *should* enable a history-less `match` on router, since we only need those methods for things like `<Link>` on the server.

This approach is definitely a work-in-progress. I'm not precious about it, so feel free to rip it apart. At the very least, this might inspire some good changes to the library.

The weirdest part is the changes to `useBasename` and `useQueries` so that they are backwards compatible. Kinda fugly, but that comes with the BC contract we're trying to maintain. I didn't want to split it into two APIs (an enhancer for location providers and another for history) so that the surface area doesn't grow and users can't create inconsistent instances. I'm most ambivalent about this (second only to the name "location provider"), so it's ripe for change.

Does this approach seem sensible? Anything obvious that I missed? Will this be tenable for what we're trying to do with `match` on router?


mjackson:
Hey @timdorr- please take a look at the work I just pushed in f12d632691ebef652d0cf5dd67a033e87746fc4a. I separated out interfacing with the various underlying history APIs into several "protocol" modules. See e.g.

- https://github.com/mjackson/history/blob/f12d632691ebef652d0cf5dd67a033e87746fc4a/modules/BrowserProtocol.js
- https://github.com/mjackson/history/blob/f12d632691ebef652d0cf5dd67a033e87746fc4a/modules/HashProtocol.js

I believe there's some overlap here with your "location provider" concept. The protocol modules are just basically a bag of low-level methods that interface with some history implementation.

I want to eventually eliminate the use of `createLocation` from the high-level server-side router API, and I think we're almost there. Please let me know if you see any glaring API warts on master.

timdorr:
IIRC (sorry, I'm on mobile), we really only need createHref for `match` to render Links, so that should be workable (I know you want to kill that API like soooooooo badly :smile:). 

Can't review the changeset at the moment, but it sounds like it supercedes this one. I'll close this out so we can work on one solution.

taion:
@mjackson 

The intent here was actually something of the opposite. `getCurrentLocation` (maybe better to call this something like `makeCurrentLocation`? – given that it's actually constructing the current location rather than just pulling it out of some variable) makes history objects have a mock-stateful API.

The intent of a "location provider" is that it's actually an entirely stateless thing that pretty much only supports `createHref` and `createLocation`. This cleans up server-side rendering, because it mostly doesn't make sense to enable `push` or `replace` on the server side – you instead only want the stateless shim that lets you construct URLs.

taion:
w/r/t dropping `createLocation` for SSR in React Router, this gets us in a good spot, but IMO it'd be better to branch on https://github.com/mjackson/history/commit/1b4003f91cfb09c47122831633226a657c5ea686 and cut a 2.1.0 that just exposes `getCurrentLocation`. That would let us clean up `match` in React Router without having to do a major version bump.

PR #233, opened 2016-02-18
Update babel-eslint dep

narr:
after npm install and npm run lint causes "TypeError: Cannot read property 'visitClass' of undefined".(https://github.com/babel/babel-eslint/issues/243) So updated it with the newest version.


taion:
The correct version for us to use is 4.x, not 5.x.

PR #235, opened 2016-02-18
set 'location.pathname' to '/' only when 'location.basename' is not '/'(fixes #234)

### LINKED ISSUE #234: When using base tag, react-router and redux-devtools, Uncaught SecurityError: Failed to execute 'pushState' on 'History' occurs
When base tag is used like "http://localhost:3000/", in addBasename function from useBasename.js 
location.basename is set to '/' and location.pathname is also set to '/' if the location.pathname is '/'.
The problem is that when it reverts to the path '/' by redux-devtools with react-router, it uses locationBeforeTransitions object which is the location object from history and the path becomes '//' in finishTransition function from createBrowserHistory.js and it cause an error like below at window.history.pushState(historyState, null, path) in finishTransition function.

Uncaught SecurityError: Failed to execute 'pushState' on 'History': A history state object with URL 'http:' cannot be created in a document with origin 'http://localhost:3000'.

I guess location.pathname should be set '/' only when location.basename is not '/' so that the path in finishTransition function will be '/'.
screen shot 2016-02-18 at 3 52 30 pm

narr:
ref - #234

### DIFF
--- a/modules/useBasename.js
+++ b/modules/useBasename.js
@@ -24,7 +24,7 @@ function useBasename(createHistory) {
           location.pathname = location.pathname.substring(basename.length)
           location.basename = basename
 
-          if (location.pathname === '')
+          if (basename !== '/' && location.pathname === '')
             location.pathname = '/'
         } else {
           location.basename = ''
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "assert": "1.3.0",
     "babel": "^5.4.7",
     "babel-core": "^5.4.7",
-    "babel-eslint": "^3.1.23",
+    "babel-eslint": "^5.0.0",
     "babel-loader": "^5.0.0",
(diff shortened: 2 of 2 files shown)


taion:
This is not the correct fix. I don't fully understand the issue you're describing, but there's no reason this should only apply when the basename is exactly `/` and not just any string that ends with `/`.

narr:
If location.basename and location.pathname end with '/', it cause "Uncaught SecurityError: Failed to execute 'pushState' on 'History': A history state object with URL 'http:' cannot be created in a document with origin 'http://localhost:3000'" as the final path in finishTransition function becomes '//'
screen shot 2016-02-18 at 11 58 16 pm
window.history.pushState(historyState, null, "//") makes the error. It happens when using redux devtools's revert(like undo) with base tag.

PR #239, opened 2016-03-03
Fix history listener being called twice in Safari, basename being prepended twice

### LINKED ISSUE #238: The `useBasename` enhancer makes the history listener getting called twice initially in Safari
This has made our analytics track double pageviews for a little while now.

Here's a jsbin with a repro case:
https://jsbin.com/renopeqava/1/edit?html,js,console

I'm seeing this on Safari 9.0.3 / Mac OSX, but from the reports I'm getting it seems to be an issue with Safari on iOS too. Seems to work fine in other browsers.

matthetherington:
Fixes #238 , also:

useBasename.js ~L8 used to look like: 

```
let { basename, ...historyOptions } = options
let history = createHistory(historyOptions)
```

but was changed to 

```
const history = createHistory(options)
let { basename } = options
```

in this commit: https://github.com/mjackson/history/commit/816318f5b3f52153dfca3ba67bd55d4ee2858614

The two aren't equivalent, was this intentional? It's broken for me - causes the basename to prepended twice. Eg, my basename is `/dealer/platform/` and with the change above I'm taken to `/dealer/platform/dealer/platform/`

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,8 @@
+## HEAD
+> Mar 3, 2016
+
+- **Bugfix:** Fix history listener being called twice in Safari, basename being prepended twice ([#238])
+
 ## [v2.0.0]
 > Feb 4, 2016
 
--- a/modules/createBrowserHistory.js
+++ b/modules/createBrowserHistory.js
@@ -49,7 +49,7 @@ function createBrowserHistory(options={}) {
 
   function startPopStateListener({ transitionTo }) {
     function popStateListener(event) {
-      if (event.state === undefined)
+      if (event.state === undefined || event.state === null)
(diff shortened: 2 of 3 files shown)


taion:
I don't think we should change that enhancer – what's most likely happening is that you have `useBasename` set up twice.

Also, the condition should just be `== null`, and it'd be good to have a test for this.

tarun:
Same symptoms - different issue (all browsers).
`react-router/userRouterHistory` automatically adds `useBasename`..
so passing in `useRouterHistory(useBasename(createHistory))` - will end up binding basename twice.

taion:
@matthetherington @tarun That's correct – the current setup requires you to just pass in `basename` as an arg to `createHistory`, rather than re-wrapping.

PR #248, opened 2016-03-18
Pin query-string to avoid prototype issues

taion:
Fixes https://github.com/mjackson/history/issues/247

Note this is against the v2.x branch and needs to be released as 2.0.2 to fix things for current users.


taion:
This is the wrong fix. Should be using `Object.prototype.hasOwnProperty.call` instead.

PR #250, opened 2016-03-18
Clean up cruft in useQueries

taion:


modules/useQueries.js
```diff
@@ -41,7 +27,7 @@ const useQueries = (createHistory) =>
         return location
 
       if (location.query == null)
-        location.query = parseQueryString(location.search.substring(1))
+        location.query = parseQueryString(location.search)
```
taion: `query-string` handles leading `?`. This is breaking for legacy `qs` users, but I don't think we need to make that accommodation here.

mjackson:
Can we get a test around this? Seems odd that none of our tests are breaking, no?

taion:
Around what specifically? Query decoding from search is already covered here. It's intentional that this is non-breaking – that `substring(1)` operation is no-op when using `query-string`.

taion:
I changed my mind. The substring change accomplishes nothing useful; all it does is make life harder for users of `qs`. I've updated this PR to just take out the old warning for nested objects.

PR #251, opened 2016-03-18
Fix usage of hasOwnProperty

taion:


taion:
Opened against wrong branch.

PR #253, opened 2016-03-18
Fix usage of hasOwnProperty

### LINKED ISSUE #247: query-string@3.0.2 causes `[TypeError: object.hasOwnProperty is not a function]`
This is a heads up that the dependency `query-string` at version 3.0.2 has a different return value for the parse method than 3.0.1 did - https://github.com/sindresorhus/query-string/issues/50.  Specifically, it causes line https://github.com/mjackson/history/blob/master/modules/useQueries.js#L13 to throw `[TypeError: object.hasOwnProperty is not a function]`.

It appears that calling `hasOwnProperty` on the object returned from `parse` will not be supported going forward (since a qs param could be named `hasOwnProperty`).

taion:
Fixes #247.

This is the correct fix regardless of what `query-string` does upstream.

We need to merge this and release a new v2.0.2 unless `query-string` cuts a new release.

### DIFF
--- a/modules/useQueries.js
+++ b/modules/useQueries.js
@@ -14,7 +14,7 @@ const defaultParseQueryString = parse
 
 function isNestedObject(object) {
   for (const p in object)
-    if (object.hasOwnProperty(p) &&
+    if (Object.prototype.hasOwnProperty.call(object, p) &&
         typeof object[p] === 'object' &&
         !Array.isArray(object[p]) &&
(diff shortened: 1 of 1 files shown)


mjackson:
Looks like we don't need to cut a new release though because query-string cut a 3.0.3 that fixes this issue.

taion:
Yup, thanks!

PR #262, opened 2016-04-02
Fix handling of <base> without explicit href

taion:


taion:
This addresses https://github.com/mjackson/history/issues/94#issuecomment-204572680 but does not otherwise resolve https://github.com/mjackson/history/issues/94.

cc @loganfuller

PR #265, opened 2016-04-04
Get the navigator property from window

jacobrask:
This simplifies mocking the window object when running server side tests.

In my particular use case this patch means that I can use React Router's `Link` by just mocking `global.window` with jsdom in Node. The alternatives are more complex mocks and require hooks, or assigning all/more properties from jsdom's `window` to the Node `global` object.


taion:
Looks like we're using `window.*` elsewhere, so this is just more consistent, too. Thanks!

PR #267, opened 2016-04-05
Attempt to fix https://github.com/reactjs/react-router/issues/3241

slorber:


modules/createBrowserHistory.js
```diff
@@ -39,7 +39,7 @@ function createBrowserHistory(options={}) {
       key = history.createKey()
 
       if (isSupported)
-        window.history.replaceState({ ...historyState, key }, null, path)
+        window.history.replaceState({ ...historyState, key }, null, window.location.href)
```
taion: The 3rd arg here should just be omitted, e.g.

```js
window.history.replaceState({ ...historyState, key }, null)
```

slorber: thanks yes I think it should work too, I'll try tomorow and edit my PR

slorber:
@taion I've tested and yes it seems to work fine for my usecase with your suggestion so I've updated the PR

slorber:
Just also want to point out that this code:

```
    // Automatically use the value of <base href> in HTML
    // documents as basename if it's not explicitly given.
    if (basename == null && _ExecutionEnvironment.canUseDOM) {
      var base = document.getElementsByTagName('base')[0];

      if (base) basename = _PathUtils.extractPath(base.href);
    }
```

It does not work to automatically solve my usecase so I have to configure manually the basename. Anyway not sure there's anything clever to be done here. Maybe history could automatically detect that basename has a different domain than current url and issue a meaningful error or use it as default basename?

I tried `if (base) basename = base.href;` but it's not really what we want here because actually the history basename should be set to the backend server for solving the security issue, while my page has base=frontend server.

taion:
See the recent discussion on https://github.com/mjackson/history/issues/94. The `<base href>` support in this library is just not correct and needs to be replaced entirely (e.g. with `<base basename>` or something).

taion:
From https://github.com/mjackson/history/issues/94#issuecomment-204649227 onward.

slorber:
@taion @mjackson @ryanflorence I've explained my issue more in-depth in https://github.com/reactjs/react-router/issues/3241#issuecomment-209970553

Now I don't want to rush you guys but I have to put something in production in 8 days :) so I need to know if you will consider merging this fix and release 2.0.2 soon, or if I should consider publishing my own fork at least temporarily

Thanks

taion:
@slorber Would you mind adding a regression test here? The triggers for the bug are obscure enough that is a nontrivial risk of a regression later on without such a test.

slorber:
ok i'll try to check how this can be tested :) if you have ideas plz tell me

taion:
Give me a sec.

taion:
I added it: https://github.com/mjackson/history/pull/274.

slorber:
great thanks :)

PR #268, opened 2016-04-07
Update query-string major version

taion:
This gives better deduping if users (like me!) are manually installing query-string.

I'm not adding this to the v2 branch (even though I want to have this change), because this brings back the (IMO correct) `Object.create(null)` behavior, and that's technically breaking, as we surface these objects back to the user.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
   },
   "dependencies": {
     "invariant": "^2.0.0",
-    "query-string": "^3.0.0",
+    "query-string": "^4.1.0",
     "warning": "^2.0.0"
   },
   "devDependencies": {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -67,7 +67,12 @@
     "location"
   ],
   "babel": {
-    "presets": [ "es2015", "stage-1" ],
-    "plugins": [ "dev-expression" ]
+    "presets": [
```
taion: This is npm's work.

mjackson:
I'm gonna fix the failing tests and get this into the v3 release. Thanks for bringing it to my attention, @taion.

PR #274, opened 2016-04-15
Add regression test for base href on separate origin

taion:


taion:
Looks like this needs to run a restricted set of browsers.

slorber:
@taion so it seems you did add my proposal change but tests still not pass :(

tell me if I can help but not sure to know how to interpret these test results. Is the problem "Some of your tests did a full page reload!" ?

taion:
That's correct – that will fail a test in Karma.

mjackson:
Looks like progress here has stalled. Closing unless someone wants to pick this up and run with it.

slorber:
As far as I know, our dev env which use base href does not work on Safari because of this issue, and we still didn't find any workaround

Issue #249, opened 2016-03-18
add support for hashes without leading slash


zdila:
`createHashHistory` could have an option to prevent leading slash in the hash.

Instead of `index.html#/foo` the URL would then be `index.html#foo`.

taion:
Might be worth thinking about this in conjunction with https://github.com/mjackson/history/pull/131 and https://github.com/reactjs/react-router/issues/601. It sounds like we need some flexibility in supporting various hash URL schemes to both offer appropriate legacy support and allow interop with e.g. Auth0.

mjackson:
That's not a bad idea, @zdila, but it introduces extra complexity for not a lot of gain IMO. I think I'd rather leave it as it is now unless there's a stronger argument than aesthetics.

zdila:
Aesthetics is a strong argument nowadays :-). In any case we'll try to implement our own HashProtocol then because feature we request is a critical issue for our customer.

PR #278, opened 2016-04-23
added hashbang support

tvervest:
I've added an option `hashbang` to the `createHashHistory` method to allow for optional hash-bang (`#!`) urls. This is part of a fix for reactjs/react-router#601. See the updated documentation for details.

### DIFF
--- a/docs/HashHistoryCaveats.md
+++ b/docs/HashHistoryCaveats.md
@@ -16,6 +16,11 @@ let history = createHistory({
 let history = createHistory({
   queryKey: false
 })
+
+// Enable hashbang URLs
+let history = createHistory({
+  hashbang: true
+})
 ```
 
 One other thing to keep in mind when using hash history is that you cannot also use `window.location.hash` as it was originally intended, to link an anchor point within your HTML document.
--- a/docs/HashHistoryCaveats.md
+++ b/docs/HashHistoryCaveats.md
@@ -19,8 +19,14 @@ let history = createHistory({
 
 // Use custom URL transformations, for example for hashbang support
 let history = createHistory({
-  transformPath: (path) =>
-    path.indexOf('!') !== 0 ? `!${path}` : path
+  transformPath: (path, encode) => {
+    // if encoding, we transform the path to our hash value
+    if (encode)
+      return path.indexOf('!') !== 0 ? `!${path}` : path
+
+    // when decoding, we transform the hash value back into the original path
+    return path.substring(1)
+  }
 })
 ```
 
--- a/docs/HashHistoryCaveats.md
+++ b/docs/HashHistoryCaveats.md
@@ -17,9 +17,10 @@ let history = createHistory({
   queryKey: false
 })
 
-// Enable hashbang URLs
+// Use custom URL transformations, for example for hashbang support
 let history = createHistory({
-  hashbang: true
+  transformPath: (path) =>
+    path.indexOf('!') !== 0 ? `!${path}` : path
 })
 ```
 
--- a/modules/createHashHistory.js
+++ b/modules/createHashHistory.js
@@ -30,13 +30,13 @@ const createHashHistory = (options = {}) => {
     transformPath = ensureSlash
 
   const getCurrentLocation = () =>
-    HashProtocol.getCurrentLocation(queryKey)
+    HashProtocol.getCurrentLocation(queryKey, transformPath)
 
   const pushLocation = (location) =>
-    HashProtocol.pushLocation(location, queryKey)
+    HashProtocol.pushLocation(location, queryKey, transformPath)
 
   const replaceLocation = (location) =>
-    HashProtocol.replaceLocation(location, queryKey)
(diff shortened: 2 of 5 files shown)


taion:
To the extent we want this feature, I think it should be more a function for generating the path (and reversing), so it can also e.g. support people who don't want the leading slash for some reason.

tvervest:
I see your point, that should solve all (weird) edge-cases people seem to use. I'll refactor the `HashProtocol` file to use a transformation function to generate the final path, then.

taion:
You need to be able to reverse as well – get the path from the location hash.

tvervest:
Could you point me to an example/test where the reverse function would be used? I can understand the use-case in (for example) `react-router` where the reverse method would be valuable to match the routes, but I'm not quite sure where/how it would be used in the `history` module.

taion:
`getCurrentLocation` or whatever... if you're trying to discover the current pathname from the hash, you'll need to be able to strip out the bang, or whatever

tvervest:
I see, the current test cases simply didn't trigger the issue you're talking about. I've added additional test cases that do and patched the change so that they're green, however I keep running into a failed test on Firefox 43 on Win 10 where the test fails and the log shows a message `"Some of your tests did a full page reload!"`.

I can't access the browserstack log and I'm not sure what's going on. If you could add the log or the error message that would be a great help :)

taion:
It would be really nice if we could support https://github.com/mjackson/history/issues/249 here as well – i.e. use this same code path to chomp the leading slash from pathnames. I think that would be the benefit of taking functions to transform and reverse path transformations – a single codepath that can handle `#/path`, `#!/path`, and `#path` (and I guess `#!path` as well)..

mjackson:
@taion I still don't see the need for a function. All of those are simple string prefixes.

taion:
I mean the case where `pathname` is e.g. `/path` but you want the URL to be `#path` instead of `#/path`. We could be clever about omitting the leading slash in certain cases, but it would be less code in the core library to accept a function that could deal with any edge cases that arise in trying to do this.

tvervest:
@mjackson don't worry about it, I haven't really had time to think about it since I posted my message.

Initially I agreed with you, @mjackson, but after seeing the question in #249 I tend to agree with @taion. It seems like it would be a nice way to make the library more flexible outside of the default use-cases. In short, the library supports a single "default" way of formatting the url. If you want something different then you're free to implement a simple conversion method (which usually is a simple string prefix) allowing you the flexibility to make it whatever you need.

It's a bit more work for the end-user than a simple parameter, but it will allow end-users to customise the library to their exact needs in a way that is a lot easier to maintain than just adding more options, but that's just my two cents ;)

tvervest:
@mjackson P.S. I'm an idiot for not realising that you put the history guards in place. But at least the tests are green now, thanks 👍

taion:
It's not even the "general" use case that I care about – it's just that the other ask that has come up here is getting rid of the leading slash. If we can satisfy both the "no leading slash" and the "hashbang" options, then I'd be happy.

tvervest:
Well, whatever the reasoning, we both agree to implement the same end-result :) The current implementation (with a transformation function going both ways) solves all of the use-cases that need to be solved.

modules/createHashHistory.js
```diff
@@ -83,7 +87,7 @@ const createHashHistory = (options = {}) => {
   }
 
   const createHref = (path) =>
-    '#' + history.createHref(path)
+    '#' + transformPath(history.createHref(path), true)
```
mjackson: Well, if we're going to use a `transformPath` function, can we at least make the default case use it as well? i.e. the default `transformPath` takes care of prepending the `#`.

tvervest: I understand your reasoning, but after having tried to implement this I came to the realisation that this would mean it in fact wouldn't be a hash history at all anymore. It would mean users would have to specify their own version of the [getHashPath()](https://github.com/tvervest/history/blob/escaped-fragment-support/modules/HashProtocol.js#L16) method. I suggest we enforce a `#` in the URL, but everything after that is customisable by the user's transformation method.

mjackson: Sounds good. Sorry for sending you on a wild goose chase, but your suggestion makes sense.

I'll have time to to a more detailed review later tonight. Thanks, @tvervest!

tvervest: No problem, anything to help :)

mjackson:
Apologies for the delay here. I'm digging in today and will get this merged.

docs/HashHistoryCaveats.md
```diff
@@ -16,6 +16,18 @@ let history = createHistory({
 let history = createHistory({
   queryKey: false
 })
+
+// Use custom URL transformations, for example for hashbang support
+let history = createHistory({
+  transformPath: (path, encode) => {
```
taion: Separate encode/decode (or transform/restore/whatever) callbacks would be cleaner than a single callback with a boolean flag IMO.

mjackson: Agreed.

PR #282, opened 2016-04-25
Added fallback for disabled sessionStorage

jaxxreal:
Some users browse the Internet with disabled sessionStorage, and polyfill can't resolve such issue. That's the main reason for this pull request.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,13 @@
 [HEAD]: https://github.com/mjackson/history/compare/v3.0.0-2...HEAD
 [#243]: https://github.com/mjackson/history/issues/243
 
+## [v3.0.0-3]
+> Apr 25, 2016
+
+- Added fallback for disabled sessionStorage
+
+[v3.0.0-3]: https://github.com/mjackson/history/compare/v3.0.0-2...v3.0.0-3
+
 ## [v3.0.0-2]
 > Apr 19, 2016
 
--- a/modules/DOMStateStorage.js
+++ b/modules/DOMStateStorage.js
@@ -8,16 +8,37 @@ const QuotaExceededErrors = [
 
 const SecurityError = 'SecurityError'
 const KeyPrefix = '@@History/'
+const STORAGE = window.sessionStorage || (() => {
+      let storage = {}
+
+      return {
+        removeItem(key) {
+          delete storage[key]
+          storage.length = Object.keys(storage)
+        },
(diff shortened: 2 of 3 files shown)


taion:
Are you sure this works? I thought disabled session storage most often led to quote exceeded exceptions, not `window.sessionStorage` not being available.

jaxxreal:
Yeah, I'm sure. In my case users use specific browsers or add-ons which define `window.sessionStorage` as `null`, and for such case my solution works as well.

taion:
I'm not opposed to this in principle as a fallback for e.g. Safari private mode where there is no session storage period, but the fallback should be broader, and we should have tests.

jaxxreal:
Ok, I will add test case for nulled sessionStorage.

taion:
Session storage has different persistence semantics, and in fact they are the correct ones for history state storage.

jaxxreal:
How about cross-browser compatibility? It's correct, but what does it mean if part of users of end product can't use it in some cases?

taion:
https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage#Browser_compatibility

jaxxreal:
Sorry, but your link don't answer on my second question.

taion:
Can you elaborate on which browsers you are working with that do not provide session storage at all?

jaxxreal:
FF 45.0 with `dom.storage.enabled = false` which set over `about:config` page.

taion:
Interesting. It's not clear to me that we get much out of polyfilling session storage in this case. It's probably sufficient to just guard against the non-existence of session storage before attempting to use it.

jaxxreal:
I don't ask you to do something else than just apply my request. It works and pass all yours test cases. If problem in another moment, could you clarify me?

taion:
I'm saying that the fallback for `window.sessionStorage` not being available should probably be to just return early in both of those methods, rather than trying to polyfill.

mjackson:
We can't polyfill `window.sessionStorage` with a hash in memory. It's not the same thing.

taion:
@mjackson I agree there, but it seems not ideal to have an uncaught exception if session storagei s disabled. Doesn't that kill the transition entirely?

jaxxreal:
I can't got your position in this question, guys. Why don't provide history functionality for browsers with disabled sessionStorage? I don't say about ie6, but about modern ones with stricter security options.

PR #288, opened 2016-05-04
namespace global name

kentcdodds:
So it doesn't clash with the built-in [History](https://developer.mozilla.org/en-US/docs/Web/API/History) api.

I'm sure everyone will hate this. I'm just not sure what to name it, so this is a place to have a discussion about what to call it.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -3,7 +3,7 @@ const webpack = require('webpack')
 module.exports = {
 
   output: {
-    library: 'History',
+    library: '_History',
     libraryTarget: 'umd'
   },
 


mjackson:
Just wanted to clarify that this is only an issue when you're using the UMD build outside of any module system (i.e. using globals).

AFAICT there's no real use case for `window.History` beyond `window.history instanceof window.History`. Not saying we should definitely clobber `window.History`, but I would like to list out the pros and cons here so we can tell if there's a good reason not to.

tleunen:
`HISTORY`?
`hstory`?

taion:
`React`

taion:
(JOKE, don't kill me)

mzabriskie:
`Herstory`
`Theirstory`

mjackson:
You forgot `Ourstory`

mjackson:
`NeverEndingStory`

taion:
`__SECRET_HISTORY_DO_NOT_USE_OR_YOU_WILL_BE_FIRED`

kentcdodds:
Here's the first decision we need to make:

> Is it worth avoiding overriding the existing `History` global?

If not, then we can close this. If so, then the next decision is:

> Does it really matter what the global is called?

IMO, it really doesn't (within reason) because the only people using this will be using it as a global (ie, not a real application) and therefore can be just about anything that's reasonable. So I'm actually good with `_History` or perhaps `mjHistory` (as in `Michael Jackson History`: A biography about the man, the myth, the legend).

rockerest:
Coming here from [this tweet thread](https://twitter.com/kentcdodds/status/727960226146619393).

I think the best use case is: `History` is what's [documented by MDN](https://developer.mozilla.org/en-US/docs/Web/API/History) despite it being [lowercased in other places](https://developer.mozilla.org/en-US/docs/Web/API/History_API), and in the [spec](https://www.w3.org/TR/html5/browsers.html#the-history-interface)([s](https://html.spec.whatwg.org/multipage/browsers.html#the-history-interface)).

My personal opinion (and I'll pretend it was requested since Kent linked me here ;P) is that the question is not:

> Is it worth avoiding overriding the existing History global?

but rather

"Is it worth intentionally creating the possibility for confusion?"

For the record, the link Kent used points to the [History API](https://developer.mozilla.org/en-US/docs/Web/API/History) while the lowercase [`window.history`](https://developer.mozilla.org/en-US/docs/Web/API/Window/history) is the interface to that object.

Personally, I think it's far more valuable to leave browser APIs intact than to assume no one would use them.

The counter to that is exactly what @mjackson said above: 
> AFAICT there's no real use case for window.History beyond window.history instanceof window.History

But - from an outside perspective - I would be concerned that forever into the future, your documentation would have to have a little asterisk that says: "Oh also we clobbered the History API because we assumed you'd never need it."

I wouldn't want to take the mantle of that responsibility.

mjackson:
@rockerest There are no methods on `window.History`. It's an interface for creating `history` instances.

taion:
And one you can't use, at that – you can't do `History()` or `new History()`.

rockerest:
@mjackson Certainly. I understand that (and did a bit of testing to make sure I understood how the parts work / interact, too).

I think the point I didn't make very well was:
"Is not having methods an okay reason to take over a namespace?"

I wouldn't personally be comfortable with the precedent this would set, but that doesn't mean it's wrong to do so.

You do you, in other words :D

mjackson:
I think we'll leave this as-is for now. When it actually breaks someone's app, we'll change it. But I don't ever see that happening.

taion:
```js
invariant(window.history instanceof window.History, 'help my app broke')
```

Done 😄

PR #293, opened 2016-05-23
call getCurrentIndex before reassigning currentLocation

maisano:
calls to `getCurrentIndex` always return `-1` because `currentLocation` is being reassigned before lookup. this fixes behavior for restoring the url on canceled transitions (without a refresh).


maisano:
closing–will open a more thorough pull request for this fix.

PR #296, opened 2016-05-24
Fix #295: Hash history fails to save state on IE 11 when served over file protocol

### LINKED ISSUE #295: Hash history fails to save state on IE 11 when served over file protocol
IE 11 does not expose a `window.sessionStorage` object when the page is served over the file protocol. The causes the `saveState` method in `DOMStateStorage` to fail. Since creating the history with `query=false` was deprecated in version 3.0.0, I am unable to use hash history in this scenario.

I noticed similar scenarios in Chrome/Firefox/Safari throw security exceptions which are handled, and the failure is simply logged. Can we add a similar behaviour if `window.sessionStorage` is undefined?

lucaas:
`window.sessionStorage` is undefined in Internet Explorer when served via file protocol.
Ignore the error and return undefined in readState/setState.

Fixes #295

### DIFF
--- a/modules/DOMStateStorage.js
+++ b/modules/DOMStateStorage.js
@@ -13,6 +13,16 @@ const createKey = (key) =>
   KeyPrefix + key
 
 export const saveState = (key, state) => {
+  if (!window.sessionStorage) {
+    // Session storage is not available or hidden.
+    // sessionStorage is undefined in Internet Explorer when served via file protocol.
+    warning(
+      false,
+      '[history] Unable to save state; sessionStorage is not available'
+    )
+    return
+  }
+
   try {
     if (state == null) {
       window.sessionStorage.removeItem(createKey(key))
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,9 +2,12 @@
 
 - `location.query` has no prototype
 - Warn about protocol-relative URLs ([#243])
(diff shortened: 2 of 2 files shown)


modules/DOMStateStorage.js
```diff
@@ -20,6 +20,17 @@ export const saveState = (key, state) => {
       window.sessionStorage.setItem(createKey(key), JSON.stringify(state))
     }
   } catch (error) {
+    if (!window.sessionStorage) {
```
taion: Check this before attempting to write... no point in catching this as an exception

modules/DOMStateStorage.js
```diff
@@ -50,6 +61,17 @@ export const readState = (key) => {
   try {
     json = window.sessionStorage.getItem(createKey(key))
   } catch (error) {
+    if (!window.sessionStorage) {
+      // Session storage is not available or hidden.
+      // sessionStorage is undefined in Internet Explorer when served via file protocol.
+      warning(
+        false,
+        '[history] Unable to save state; sessionStorage is not accessible'
```
taion: this is attempting to "read", not "save"

lucaas:
I've now updated it to check before trying to access the object and removed the readState check, since it doesn't re-throw the error.

mjackson:
Thanks, @lucaas! :D

PR #209, opened 2016-01-15
Fix replacing hash path in various edge cases

taion:
Fixes #126

PR #297, opened 2016-05-24
Fix #126: Replacing hash path in IE served via file protocol

### LINKED ISSUE #126: HashtagHistory on Filesystem not working with IE
Hi

This occurs when I open a react website on the filesystem (e.g. index.html) using HashHistory.
##### Firefox, Chrome, Safari
Opening the file `X:\index.html` results in the following:
1. Firefox opens and loads the URL `file:///X:/index.html`
2. The webpage gets loaded fine.
3. The URL gets rewritten to `file:///X:/index.html#/?_crypticid`

##### IE11
Opening the file `X:\index.html` results in the following:
1. IE11 opens and loads the URL `file:///X:/index.html` resp. `X:\index.html`
2. The webpage displays briefly a blank page.
3. The URL gets rewritten to `X:\X:\index.html`
4. IE11 redirects to the IE internal error page "dnserror.html".

In IE11 following messages are shown in the console:
Information: `HTML1300: Navigation occurred.`
Warning: `DOM7011: The code on this page disabled back and forward caching. For more information, see: http://go.microsoft.com/fwlink/?LinkID=291337`

The same site gets loaded just fine in IE, when I use `createMemoryHistory` instead of `createHashtagHistory`.

Any help would be greatly appreciated :)

lucaas:
Fixes the issue discussed in #126, where the initial `replaceHashPath` call replaces the current location with a bad location, when the page is served via the the file protocol in IE 11. 

The behaviour *should* be the same as before for other cases.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,9 +2,11 @@
 
 - `location.query` has no prototype
 - Warn about protocol-relative URLs ([#243])
+- **Bugfix:** Fix replacing hash path in IE served via file protocol ([#126])
 
 [HEAD]: https://github.com/mjackson/history/compare/v3.0.0-2...HEAD
 [#243]: https://github.com/mjackson/history/issues/243
+[#126]: https://github.com/mjackson/history/issues/126
 
 ## [v3.0.0-2]
 > Apr 19, 2016
--- a/modules/HashProtocol.js
+++ b/modules/HashProtocol.js
@@ -26,7 +26,7 @@ const pushHashPath = (path) =>
 
 const replaceHashPath = (path) =>
   window.location.replace(
-    window.location.pathname + window.location.search + '#' + path
+    '#' + path
   )
 
 const ensureSlash = () => {


taion:
Why does this pass CI while #209 did not?

taurose:
Pretty sure it's because of 2fe8fa04aba27e2e21272859bd5083f60601f789 . As I said in [this comment](https://github.com/mjackson/history/issues/126#issuecomment-172677889), a `base` tag causes this to fail by affecting the resolution of `#foo`. For example, on a page with the url `/` and a base tag set to `/bar`, `#foo` would resolve to `/bar#foo` instead of `/#foo`, causing a page reload.

taurose:
The main drawback I can think of are use cases like reactjs/react-router/issues/3241 where base tag is intended to affect the URLs of assets. One can work around that with `useBasename`, but that's not possible here since hash history works differently.

I think ideally we'd want hash history to ignore the base url in order to never leave the page, i.e. something like a cross-browser `new URL('#' + path, document.URL).href`.

How about

```js
  const i = window.location.href.indexOf('#')

  window.location.replace(
    window.location.href.slice(0, i >= 0 ? i : 0) + '#' + path
  )
```

taion:
Setting basename from base href is gone in 3.x, no?

taurose:
Yes, I'm referring to cases where people want to use base tag independent of history and routing like https://github.com/reactjs/react-router/issues/3241 .

With this PR, you can't use `<base href>` at all in combination with hash history.

taion:
Ah, got it. Probably we should add those test cases back.

mjackson:
Thanks for the discussion here, guys. @lucaas What do you think of @taurose' idea in https://github.com/mjackson/history/pull/297#issuecomment-221715665 ?

lucaas:
@mjackson, @taurose Good catch, I've changed to your approach. I've verified that new code works for my use case and don't see any other issues with it. 

Are you able to run and add back the test cases you discussed?

mjackson:
Thanks, @lucaas :D I merged manually and resolved conflicts.

PR #300, opened 2016-05-25
Fix installation on windows

taurose:
`build-umd` and `build-min` are run when source package is installed as a result of the `prepublish` hook, causing errors on windows.

Solution is the same as reactjs/react-router/pull/3383


taion:
LGTM

mjackson:
Thanks for the PR, @taurose. Instead of wrapping our executable inside `cross-env` (which spawns another process) let's just pass in our own `env` vars in our build script. I'll push a commit to show you what I mean.

mjackson:
This PR also highlighted another issue which I've been meaning to fix for a while, which is that our build runs on `npm install`. I fixed this in c21b426d7de01faf4bc1aed67866abcab643a53a.

taion:
BTW, as on React Router, we'll still need to use the `BABEL_ENV` stuff to fix the ES modules build.

That said, I'd seriously consider dropping that build for the time being. Like... there's a tiny benefit to module size from not having the transpiled Babel junk, maybe, but tree-shaking doesn't really work that well in practice most of the time, so we can't really change import style throughout, which was sort of the point of that failed exercise.

mjackson:
The only reason we have that build is because you made it. I'll happily
drop it.

taion:
I think we should. I was afraid you'd decided to keep it given that you haven't dropped it yet!

I thought at the time that tree shaking could replace deep imports, but it can't. So given that and the difficulty of making an ES module build with Babel 6 means that we should drop that build on major version boundaries.

mjackson:
:thumbsup:

PR #301, opened 2016-05-31
Bugfix: restore previous URL on canceled popstate transitions

maisano:
reassigning `currentLocation` prior to getting the current index ensured that `-1` was always returned. this broke behavior for restoring the previous url on canceled popstate transitions.

this pull request fixes that behavior and adds a new test to verify the fix.


maisano:
thoughts? i must assume this is pretty low priority, but it would be nice to see it fixed.

mjackson:
That makes sense. Thanks, @maisano! :D

PR #306, opened 2016-06-07
Docs: remove reference to base href

MoOx:
You remove `<base>` support since v3 https://github.com/ReactJSTraining/history/blob/master/CHANGES.md#v300-1 so I guess this should be remove in doc too.


taion:
Thanks!

PR #314, opened 2016-06-16
Update createHistory.js

dckesler:
When using this library, my coworker and I ran into a problem with the location being stale. After creating history we would need to turn off the event listeners, and then at a later point re-enable the history and event listeners. Unfortunately, when the listen function ran after re-enabling history `location` was the url of when we disabled history rather than being reflective of the current url. By running `getCurrentLocation` on each listen the location is always current.

### DIFF
--- a/modules/createHistory.js
+++ b/modules/createHistory.js
@@ -70,13 +70,20 @@ function createHistory(options={}) {
 
   function listen(listener) {
     changeListeners.push(listener)
-    
-    const location = getCurrentLocation()
-    allKeys = [ location.key ]
-    updateLocation(location)
+
+    if (location) {
+      listener(location)
+    } else {
+      const location = getCurrentLocation()
+      allKeys = [ location.key ]
+      updateLocation(location)
+    }
 
     return function () {
       changeListeners = changeListeners.filter(item => item !== listener)
+      if (changeListeners.length === 0) {
+        location = undefined
+      }
     }
   }
 
--- a/modules/createHistory.js
+++ b/modules/createHistory.js
@@ -70,14 +70,10 @@ function createHistory(options={}) {
 
   function listen(listener) {
     changeListeners.push(listener)
-
-    if (location) {
-      listener(location)
-    } else {
-      const location = getCurrentLocation()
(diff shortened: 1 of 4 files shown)


taion:
This is definitely not the right thing to do if the history already has other listeners attached.

Why not just blank out the current location when the last listener is removed?

dckesler:
We tried to make the change more along the lines of what you said. Not sure why the `browser history "before each" hook for "works"` test on Safari 9 is failing though.

mjackson:
@dckesler Do you have any time to work up a failing test that illustrates the stale location problem? That would be invaluable here.

dckesler:
Sure, I'll try and get it done tomorrow.

taion:
Any update here?

joeldenning:
+1 I'm interested in seeing the tests get added so that this can get merged and published.

taion:
Ping! It'd be good to get some sort of regression test just to avoid breaking this again.

dckesler:
Obviously we didn't get around to as soon as I had expected, but this last commit includes tests that fail without our change.

joeldenning:
Any update on this? Does anybody know what the document exception 13 means in safari that's causing the tests to be broken? We've been using @dckesler's fork to be able to support react-router inside of several production single-spa applications (https://github.com/CanopyTax/single-spa), and would love to see this merged and published. The fix in this pr solves a problem where react router apps that are mounted, unmounted, and then remounted have a stale location.

Appreciate all that you guys do!

taion:
It's a dumb Safari thing. Unrelated to this PR.

modules/createHistory.js
```diff
@@ -81,6 +81,9 @@ function createHistory(options={}) {
 
     return function () {
       changeListeners = changeListeners.filter(item => item !== listener)
+      if (changeListeners.length === 0) {
+        location = undefined
```
taion: Make this `null` instead of `undefined` and I'll merge.

Going to follow up w/@mjackson on getting a release out.

dckesler:
@taion Made the change

taion:
Thanks.

PR #315, opened 2016-06-20
Fix dev-dependencies

taion:
babel-core is a peer dependency of babel-loader, so this squelches the warning.

The second commit bumps up all dep ranges to the latest version, just as a housekeeping detail.


mjackson:
Thanks, @taion! :D

PR #316, opened 2016-06-20
Fix key for new locations with hash history

taion:
The new test case demonstrates the issue. When you do `hashHistory.push()`, the new location has a `key`, but there's no query key and you can never recover that key again; it ends up being a purely in-memory thing.

This code isn't wonderful but it's the best I could find without a more extensive refactor.

### DIFF
--- a/modules/__tests__/describeQueryKey.js
+++ b/modules/__tests__/describeQueryKey.js
@@ -2,7 +2,7 @@ import expect from 'expect'
 import { PUSH, POP } from '../Actions'
 import execSteps from './execSteps'
 
-const describeQueryKey = (createHistory) => {
+const describeQueryKey = (createHistory, getFullPath) => {
   describe('when queryKey == "a"', () => {
     let history
     beforeEach(() => {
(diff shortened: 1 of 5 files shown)


modules/__tests__/describeQueryKey.js
```diff
@@ -48,6 +51,19 @@ const describeQueryKey = (createHistory) => {
...
+          expect(getFullPath()).toEqual(`/home?the=query&a=${location.key}`)
+
+          history.push('/')
+        },
+        (location) => {
+          expect(location.pathname).toEqual('/')
+          expect(location.search).toEqual('')
+          expect(location.state).toBe(undefined)
+          expect(location.action).toEqual(PUSH)
+          expect(location.key).toBe(null)
```
taion: This previously would not have been `null`.

taion:
I'm not sure how much this matters given other pending work, but this addresses my biggest concern in https://github.com/ReactJSTraining/history/issues/163#issuecomment-201895122 – namely that `location.key` on a hash history is _not_ `null` on `PUSH` transitions.

taion:
Rebased

taion:
Ping!

mjackson:
@taion You're welcome to cut a 3.x maintenance release with this work in it. The majority of the work for v4 is already done, and I'm focused on shipping that for the time being.

mjackson:
Looks like this has stalled. Please re-open if you'd still like to get this merged into a 3.x maintenance release.

PR #318, opened 2016-06-24
Fix object equality check

mruoss:
The check ```statesAreEqual(a,b)``` currently returns ```true``` for ```a={x: 1}``` and ```b={x: 1, y: 2}```.

### DIFF
--- a/modules/LocationUtils.js
+++ b/modules/LocationUtils.js
@@ -48,9 +48,12 @@ export const statesAreEqual = (a, b) => {
       'You must not store Date objects in location state'
     )
 
-    if (!Array.isArray(a))
-      return Object.keys(a).length === Object.keys(b).length &&
-        Object.keys(a).every(key => statesAreEqual(a[key], b[key]))
+    if (!Array.isArray(a)) {
+      const keysofA = Object.keys(a)
+      const keysofB = Object.keys(b)
+      return keysofA.length === keysofB.length &&
+        keysofA.every(key => statesAreEqual(a[key], b[key]))
+    }
 
     return Array.isArray(b) &&
       a.length === b.length &&
--- a/modules/LocationUtils.js
+++ b/modules/LocationUtils.js
@@ -49,7 +49,8 @@ export const statesAreEqual = (a, b) => {
     )
 
     if (!Array.isArray(a))
-      return Object.keys(a).every(key => statesAreEqual(a[key], b[key]))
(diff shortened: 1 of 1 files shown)


taion:
LGTM

taion:
Actually, cache the values of `Object.keys(a)` so you don't have to call it twice.

mruoss:
You mean by defining a const inside the if-stmt?

taion:
Yes.

mruoss:
better now?

taion:
Thanks.

BTW, the semantics here aren't the same as a standard deep equality comparison – it's possible that `a` and `b` could have disjoint key sets of the same size, but that `b` gets a key from its prototype that `a` has directly.

It's not worth adding the extra logic here, but I think it'd be better to just use `isEqual` from Lodash – even Redux has a dependency on Lodash these days.

taion:
This is good to go from my end though.

mruoss:
Thanks.
But yeah I had the same thought (of using Lodash for the check). Now with peer dependencies it's not such a big deal anymore...

mjackson:
I'd be happy if you guys wanted to add lodash as a dep. Only reason I didn't do that in the first place is because I wanted to get the behavior right and I wasn't familiar enough with the lodash library to know if they have a method that did exactly what I wanted. But if they do, please feel free to use it here.

taion:
This used to be `deep-equal`, no? Was something wrong with the behavior there? Or did you specifically want the logic as implemented right now v. e.g. what's on this PR?

mjackson:
deep-equal is a huge dep. I wanted something smaller.
On Mon, Jun 27, 2016 at 10:11 AM Jimmy Jia <notifications@github.com> wrote:

> This used to be deep-equal, no? Was something wrong with the behavior
> there? Or did you specifically want the logic as implemented right now v.
> e.g. what's on this PR?
>
> —
> You are receiving this because you commented.
> Reply to this email directly, view it on GitHub
> <https://github.com/ReactJSTraining/history/pull/318#issuecomment-228810297>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAFqp8VI6Hq_6Qy42UdcKZ4KvdZpbWb7ks5qQAQ5gaJpZM4I9zyI>
> .
>
-- 
Michael Jackson
@mjackson

mjackson:
Thanks, @mruoss!

Issue #305, opened 2016-06-03
exports.Actions is undefined?


Zae:
Hi there,

I tried using your constants from Actions (which seem to be exported), but it seems that it is undefined.

Is there another way to use the same consts?

jmurzy:
How do you `import` them? This should work:

```js
import { Actions } from 'history';
const { PUSH, REPLACE, POP } = Actions;
```

Zae:
hi @jmurzy,

That's exactly what I tried. I'm getting this error.

`Uncaught TypeError: Cannot read property 'PUSH' of undefined`

taion:
This was fixed on v3 in https://github.com/ReactTraining/history/commit/b3db5e14b2b707993627833af3089ec09007376f.

Issue #302, opened 2016-05-31
createLocation is not present in your mudule any more


darkfrontcode:
Hey man, I install latest version of you module, but, createLocation is not present any more. I try any ways, like import { createLocation } from 'history', or call from history.createLocation and nothing. Can U help me? Thank you.

taion:
The top-level one has been deprecated for a long time now. It's on history objects now.

darkfrontcode:
you mean this project? https://www.npmjs.com/package/object-history or another thing?

chentsulin:
https://github.com/mjackson/history/blob/master/modules/createHistory.js#L161

darkfrontcode:
Thank you my friend =}

calvinl:
@darkfrontcode @chentsulin @taion How do we get `createLocation` for environments with no DOM? It's not present in the top level anymore.

It's also still listed in https://github.com/ReactJSTraining/history/blob/master/docs/Location.md#programmatic-creation

taion:
That's a docs bug then. Creating a location without a history object (which need not require a DOM, e.g. with memory history) has been deprecated for a while, because location creation depends on history enhancers/&c.

calvinl:
@taion thanks -- good to know!

shark0der:
For others that were wondering how to do this on server side:
```
const location = createLocation(req.url);
match({ routes, location }, (error, redirectLocation, renderProps) => {
```
becomes:
`match({ routes, location: req.url }, (error, redirectLocation, renderProps) => {`

Credits to @calvinl for referencing this issue in his commit.

PR #326, opened 2016-07-14
Fix the docs for createLocation

taion:
I know it's intended to be deprecated, but people have complained about this in e.g. #302 and #305.


taion:
Just docs, merging

Issue #291, opened 2016-05-10
iOS 9.3 limits the number of times pushState can be called


iatzmon:
Hi,
Since version 9.3 of iOS Apple has limited the number of times a page can call the pushState API to 100 times. Any further attempt causes the following exception to be thrown:

`SecurityError: DOM Exception 18: An attempt was made to break through the security policy of the user agent.`

You can read more about it [here](https://forums.developer.apple.com/thread/36650).

It seems that this is their (terrible) solution for an exploit using this API.
My app has many pages and can easily reach more than a 100 router transitions. Currently, the only workaround I can think is to catch this exception and reload the application once it happens with the requested route, which resets the pushState counter back to 0.

taion:
This seems like it might be a pretty valuable workaround to add.

How specific is this exception message?

Dattaya:
According to a comment in this bug report: https://bugs.webkit.org/show_bug.cgi?id=156115 it's 100 calls per 30 second interval, so it's not that bad as I thought initially.

amarchen:
Depends on how your app works. I was updating URL on every slider move and hit security exception after few seconds of playing with a slider.
Ember.js folks also experienced this issue and reload the page when this exception happens - http://jdurand.com/blog/2016/05/03/ember-history-pushstate-dom-exception-18/

If the critical point is indeed 100 calls per 30 seconds, could the [optional] throttling be a solution? Say, if the code is running on a platform affected (i.e. Mobile Safari), we could either hard limit speed of url update to 3.4 per second.

Or a more clever exponential formula could be possible, so that first updates could happen instantly, then nearly instantly and if code continues pushing state too quickly, speed of updates would go down so that it never goes over 100 per last 30 sec.

taion:
I'd go with the hard refresh on this exception instead of trying to be clever. I think any attempted cleverness is more likely to cause weird bugs.

Though hitting the history API >3x per second for 30 seconds seems... excessive? For most use cases.

amarchen:
I guess throttling (both stupid 3 times a sec and clever with changing number of skipped requests as we come closer to the limit) could be done through the wrappers/enhancers. Just as `useQueries()` enhances `history` with the useful fields, `iosThrottling()` could introduce a throttling mechanism. 

Not sure if it should be a part of this particular project though - I guess it depends on how much `history` project wants to be "we provide the full no-trouble history service" VS "we are just a convenient utility for the browser history".

[...]

PR #328, opened 2016-07-14
Add "with-" enhancers, deprecate "use-" enhancers

### LINKED ISSUE #322: Use withFoo(history) enhancers
@mjackson 

This came up earlier, wanted to run it by you before I code it up – I think we should switch to e.g. `withBasename(history, basename)` instead of `useBasename(createHistory)({ basename })`.

This will make it easier to enhance e.g. the default `browserHistory` coming out of React Router, and is just otherwise a simpler API.

taion:
Closes #322

### DIFF
--- a/docs/BasenameSupport.md
+++ b/docs/BasenameSupport.md
@@ -3,12 +3,10 @@
 Support for running an app under a "base" URL is provided by the `useBasename` [enhancer](Glossary.md#createhistoryenhancer) function. Simply use a wrapped version of your `createHistory` function to create your `history` object and you'll have the correct `location.pathname` inside `listen` and `listenBefore` hooks.
 
 ```js
-import { createHistory, useBasename } from 'history'
+import { createHistory, withBasename } from 'history'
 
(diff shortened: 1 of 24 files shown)


taion:
I lumped in some warning-related test setup, since I want to assert on the deprecation warnings.

You'll want `?w=1` for the diffs on the tests to make sense. They look a bit messy but it was the best way I could think of to use the same tests for both "use-" and "with-" enhancers.

taion:
The tests are probably failing because of https://github.com/ReactTraining/history/issues/291. 😑

taion:
Rebased

taion:
Any update here?

mjackson:
Thanks for taking this on, @taion, and my apologies for not getting around to it sooner.

The future of this library is to just ditch the whole `use*` pattern entirely. I wouldn't be entirely opposed if you wanted to cut a 3.x release with this work in it, but I don't see the point. Most of the work for v4 is already on its way.

PR #329, opened 2016-07-14
Use es2015-loose instead of es2015

taion:
This drops both the unminified and the minified UMD builds by 1 kB or so, and probably makes code run marginally faster.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
     "babel-plugin-dev-expression": "^0.2.1",
     "babel-plugin-transform-object-assign": "^6.8.0",
     "babel-preset-es2015": "^6.9.0",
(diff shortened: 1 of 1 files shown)


taion:
And the v2 build pipeline was doing `--loose all` anyway.

mjackson:
Excellent, thank you. I was just thinking the other day I should probably be using `es2015-loose` everywhere.

taion:
I think `es2015-loose` is generally the right preset to use.

`es2015` did help me catch a bug once where I was using the new spread operator on a `null` value, though. Loose mode is okay with it (treats it as an empty iterable), but it's supposed to throw an exception.

taion:
(array spread, not object spread)

PR #330, opened 2016-07-14
Fix basename normalization

taion:
I'm opening this against v2.x because that's where the repro is. This should be easy to port to v3.x.

This address the issue @MoOx and others reported at https://github.com/reactjs/react-router/issues/3184#issuecomment-225448418 and elsewhere.

When using a basename with a trailing slash in React Router, along with transition hooks, when you try to push a pathname of '/', the `listenBefore` hooks on the history will fire. This will call `addBasename` from the `useBasename` enhancer, which will mutate the location object in place.

This causes problems with the basename has a trailing slash. Suppose the basename is "/" and the full pathname is also "/". The logic in `addBasename` will set up the location object such that it has `{ basename: '/', pathname: '/' }`. `createBrowserHistory` at https://github.com/ReactTraining/history/blob/v2.1.2/modules/createBrowserHistory.js#L79 will compute a `path` of `//` by concatenating `basename` and `pathname`, which will cause problems.

This PR resolves this issue by not mutating the input `location` object in `addBasename` when adding the basename.

It also adds a few minor cleanups:
- Normalize the basename once, up front, instead of doing it every time in `prependBasename`
- Rearrange the code in `addBasename` a little bit to put the guard conditions first

### DIFF
--- a/modules/useBasename.js
+++ b/modules/useBasename.js
@@ -4,11 +4,20 @@ import { parsePath } from './PathUtils'
 import runTransitionHook from './runTransitionHook'
 import deprecate from './deprecate'
 
+function getNormalizedBasename(basename) {
+  // Remove trailing slash from basename.
+  if (basename && basename.slice(-1) === '/')
+    return basename.slice(0, -1)
+
+  return basename
+}
+
 function useBasename(createHistory) {
   return function (options={}) {
     const history = createHistory(options)
 
     let { basename } = options
+    let normalizedBasename = getNormalizedBasename(basename)
 
     let checkedBaseHref = false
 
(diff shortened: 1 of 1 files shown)


taion:
@MoOx I can't seem to get Phenomic to point at my custom build of history. Can you check that this resolves your issue? I think CI is failing because of #291.

mjackson:
I restarted the tests.

taion:
I think the v2.x test suite is just going to consistently fail unless we either cut down the number test cases or drop the Safari browsers. Or maybe test against older versions of Safari?

mjackson:
>  maybe test against older versions of Safari?

Sure, feel free to take that approach to get this work merged into 2.x.

mjackson:
Looks like this has stalled. Please re-open if you still want to get it merged.

PR #336, opened 2016-07-22
Add hashchange listener for IE11 to fix #335

### LINKED ISSUE #335: History event not fired on hash-only change in IE11
See https://github.com/reactjs/react-router/issues/3655 for details

jrajav:
Fixes #335

### DIFF
--- a/modules/DOMUtils.js
+++ b/modules/DOMUtils.js
@@ -1,3 +1,5 @@
+const ua = window.navigator.userAgent
+
 export const addEventListener = (node, event, listener) =>
   node.addEventListener
     ? node.addEventListener(event, listener, false)
@@ -16,8 +18,6 @@ export const removeEventListener = (node, event, listener) =>
  * changed to avoid false negatives for Windows Phones: https://github.com/reactjs/react-router/issues/586
  */
 export const supportsHistory = () => {
-  const ua = window.navigator.userAgent
-
   if ((ua.indexOf('Android 2.') !== -1 || ua.indexOf('Android 4.0') !== -1) &&
     ua.indexOf('Mobile Safari') !== -1 &&
     ua.indexOf('Chrome') === -1 &&
@@ -32,10 +32,11 @@ export const supportsHistory = () => {
  * Returns false if using go(n) with hash history causes a full page reload.
  */
 export const supportsGoWithoutReloadUsingHash = () =>
-  window.navigator.userAgent.indexOf('Firefox') === -1
+  ua.indexOf('Firefox') === -1
 
 /**
- * Returns true if browser fires popstate on hash change
+ * Returns true if browser fires popstate on hash change.
+ * IE10 and IE11 do not.
  */
 export const supportsPopstateOnHashchange = () =>
-  !navigator.userAgent.match(/Trident.*rv[ :]*11\./)
+  ua.indexOf('Trident') !== -1
--- a/modules/DOMUtils.js
+++ b/modules/DOMUtils.js
@@ -1,5 +1,3 @@
-const ua = window.navigator.userAgent
-
 export const addEventListener = (node, event, listener) =>
   node.addEventListener
     ? node.addEventListener(event, listener, false)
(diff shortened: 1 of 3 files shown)


taion:
Our test suite covers IE11 – would you mind adding a test case to ensure this has the correct behavior of firing the callback exactly once across all covered browsers?

jrajav:
Can do! Updating shortly. I'd wrongly assumed IE11 wasn't covered.

mjackson:
This looks like a great addition. Thanks, @jrajav. Would love to see a test case though. Thanks :)

jrajav:
Test added!

modules/DOMUtils.js
```diff
@@ -1,3 +1,5 @@
+const ua = window.navigator.userAgent
```
taion: This is going to be a `ReferenceError` on node

jrajav:
@taion Fixed!

modules/__tests__/describePopState.js
```diff
@@ -18,6 +18,14 @@ const describePopState = (createHistory) => {
 
       window.history.back()
     })
+
+    it('is called when only a hash is navigated to', (done) => {
+      unlisten = history.listenBefore(() => {
+        done()
```
taion: Use `listen` instead of `listenBefore`. Also, assert that the listener is called exactly once.

jrajav: Should I update the test before it with both improvements, as well?

jrajav:
I ran into some issues. Updating the previous test (the one that existed before this PR) the same way actually caused it to fail, since it was calling the listener twice. And the reason for that might be the same as the second issue I ran into, though I didn't confirm: The hash history actually interferes with the test for this PR since it sets window.location.hash in the test beforeEach, and the history instance under test picks that up and calls the listener. I added another setTimeout around the whole test to attach the listener only after that happens.

modules/__tests__/describePopState.js
taion: I don't think that's unnecessary. Technically these are two separate conditions anyway – one to check that we get correct `listenBefore` hook semantics for navigation actions triggered outside of `history`; yours is just to make sure that for a normal `listen` listener, we do fire the callback and we don't fire it twice.

jrajav:
Actually, due to the issue mentioned above, the previous test (the existing 'listenBefore' one) might actually be trivially passing. It might be good to update it after all...

taion:
I've definitely run into finicky stuff with the hash change tests cases in scroll-behavior. Could you show me the incorrectly failing test case?

modules/__tests__/describePopState.js
```diff
@@ -14,11 +14,17 @@ const describePopState = (createHistory) => {
     })
 
     it('is called when browser navigation is used', (done) => {
-      unlisten = history.listenBefore(() => {
-        done()
-      })
+      setTimeout(() => {
```
jrajav: Here is the issue - if this outer setTimeout is not in place, the spy will actually be called twice. I'm unsure if this is because of `window.location.hash = ''` on HashHistory-test.js:L19, or `history.push('/home')` on L8 of this file, but it seems likely to be the former since it interferes with the following hashchange test I added, as well. In any case, some way to bypass those first events is necessary for this test (such as this brute-force timeout), otherwise it's trivially testing that setup.

taion: That makes sense. Do you have a commit that demonstrates this failure? I'd like to take a closer look when I get a chance.

jrajav:
Unfortunately, it also seems to be kind of a guessing game as to how long the timeouts need to allow the events to get fired, yet not exceed the 2000ms test timeout when the processing time is added... Exceeding it is what caused the build to fail just now. Reducing the time on each timeout.

jrajav:
@taion This one (c7d42ff) has the outer setTimouet removed and fails the browser navigation unit test with 2 spy calls.

jrajav:
Okay - this latest commit seems to pass. Both tests are updated.

taion:
Okay, sorry for the wild goose chase here. It turns out that history ignores repeated attempts to invoke listeners with the same location anyway, so the spy there isn't actually going to catch any errors.

I submitted a PR against your PR branches to simplify down that test coverage and fix a code style nit: https://github.com/jrajav/history/pull/1.

Otherwise, this LGTM.

taion:
(I mean, with my nit changes on top 😄 )

jrajav:
Ah! Sounds good, haha.

mjackson:
Thanks, @jrajav. And thanks @taion for getting this PR through :D

PR #338, opened 2016-07-24
Fix History.createLocation definition in glossary

vslinko:


taion:
@mjackson Are those extra 2 args supposed to be part of the public API? Seems like they shouldn't be.

vslinko:
private args? this is something new 😀

taion:
No, I mean... I don't see why those should be user-accessible. I feel like those args shouldn't be part of the API. I dunno.

mjackson:
Yep, they're there right now so we may as well document them. They might not be there forever though.

PR #341, opened 2016-07-30
[fixed bug] history.createLocation dose'nt work with string inpunt

Lucifier129:


taion:
Could you add a test case?

Lucifier129:
sure

PR #346, opened 2016-08-08
[fixed] hash history trigger listeners twice when push a new location

Lucifier129:


taion:
What is this supposed to do? The guard in `transitionTo` in `createHistory` already catches this. Your test case passes against master.

Lucifier129:
@taion history.createHashHistory will trigger listeners twice when path is changed

```javascript
history.push('/test') // trigger listeners and modify location.hash 
hashchange handler // trigger listeners
```

`locationsAreEqual` at `transitionTo` can not detect this problem, because of the `key` is different between `nextLocation` and `currentLocation`.

taion:
Which browser do you see this on?

Lucifier129:
It is chrome 47.0.2526.73

taion:
Can you push up a commit with just the test case and not your code change? I want to see the CI fail on that. The test ran successfully for me locally without the code change.

Lucifier129:
OK

PR #348, opened 2016-08-10
[test-case] should trigger only once when path is changed

Lucifier129:


modules/__tests__/describePush.js
```diff
@@ -32,6 +32,17 @@ const describePush = (createHistory) => {
...
         execSteps(steps, history, done)
       })
+
+      it('should trigger only once when path is changed', (done) => {
+        const spy = expect.createSpy()
+        const unlisten = history.listen(spy)
+        history.push('/test')
+        setTimeout(() => {
+          expect(spy.calls.length).toBe(1)
+          unlisten()
```
taion: Please catch errors and ensure that you call `unlisten` even if the assertion fails.

taion:
I see now. Thanks. Do you want to move the fix onto this PR, or re-open the old one?

Lucifier129:
already done

modules/__tests__/describePush.js
```diff
@@ -32,6 +32,22 @@ const describePush = (createHistory) => {
...
+        const unlisten = history.listen(spy)
+        history.push('/test')
+        setTimeout(() => {
+          try {
+            expect(spy.calls.length).toBe(1)
+            done()
+          } catch(error) {
+            throw error
+          } finally {
+            unlisten()
```
taion: Better to just put this in an `afterEach` IMO, given that you control the test case here.

modules/__tests__/describePush.js
```diff
@@ -32,6 +32,22 @@ const describePush = (createHistory) => {
...
+          }
+        }, 10)
```
taion: Do you need the explicit 10 ms timeout here? I feel like it's not ideal to hard-code timeouts when avoidable.

modules/HashProtocol.js
```diff
@@ -61,9 +61,13 @@ export const startListener = (listener, pathCoder, queryKey) => {
       replaceHashPath(encodedPath)
     } else {
       const currentLocation = getCurrentLocation(pathCoder, queryKey)
+      const isEqualKey = prevLocation && currentLocation.key && prevLocation.key === currentLocation.key
+      const isEqualPathname = prevLocation && prevLocation.pathname === currentLocation.pathname
+      const isEqualSearch = prevLocation && prevLocation.search === currentLocation.search
 
-      if (prevLocation && currentLocation.key && prevLocation.key === currentLocation.key)
+      if (isEqualKey || isEqualPathname && isEqualSearch) {
```
taion: This logic isn't quite right. See perhaps https://github.com/ReactTraining/history/blob/f247b4001cd1ad8e477a1e131cada1cba58fc1fe/modules/createHistory.js#L100-L106.

Aben:
``` js
let history =  createHashHistory()
history.listen(location => {
  console.log(location.pathname) //  execute twice, why???
})
// Let's follow the code
push('/foo') 
// and then invoke transitonTo
const push = (input) =>
  transitionTo(createLocation(input, PUSH))
// and then 
if (pushLocation(nextLocation) !== false)
  updateLocation(nextLocation)
// in updateLocation, listener will be executed(1) 
// Let's focus on pushLocation
if (getHashPath() !== path) {
  pushHashPath(path)
}
// hashchange event is fired, next step invoke transitionTo again(2).
const pushHashPath = (path) =>
  window.location.hash = path
```

taion:
Why did you close this? This looks like a good fix to have.

Aben:
I readed again.  trigger twice only `push/replace` without `state`, so  we could add `queryKey` for all path, but it's ugly, this is a good fix. @Lucifier129 @taion

taion:
Hmm, #316 might fix this as well then.

taion:
I'm going to try to pull this test case into #316.

taion:
Done – and this is in fact fixed in #316

PR #357, opened 2016-08-30
fix: ignore URIError in useQueries's defaultParseQueryString

Tin:
address https://github.com/mjackson/history/issues/312


taion:
This is definitely not right. We should handle and surface the error, not attempt to do the parsing ourselves in a finicky and likely incorrect way.

PR #369, opened 2016-09-23
new folder structure uses /lib/, update readme

colinmegill:


mjackson:
Thanks for the PR @colinmegill, but that's actually the *old* folder structure :) You can check it out [here](https://unpkg.com/history@4.2.0/). Which version are you using?

colinmegill:
Ah! Interesting... I just npm install --save as in the readme, so assumed
it was latest. Chrome said not found, dug into the module and found it in
lib. I think the install I *thought* I did got skipped because react-router
had an older version. Thanks!

PR #373, opened 2016-10-02
fix the link of documentation

malcolmyu:
The link of v2.x's documentation is incorrect after the release of v4.x, which is inconvenient to look up the previous docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ history.push({
 unlisten()
 ```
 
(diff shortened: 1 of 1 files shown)


mjackson:
Thanks for pointing this out. I think it's probably better to just delete this link now that all the docs are in the README.

mjackson:
Oh, sorry. I didn't notice this PR was made against the 2.x branch. Thanks!

PR #374, opened 2016-10-07
Add in doc another helpful usage of createMemoryHistory that was very…

slorber:
… helpful for me

If you have a mobile website, that is also shipped as a Cordova app, and both have the same codebase, the Cordova app can use memory history, while the mobile website can use real browser history.
This way, the cordova app can stay on index.html of Cordova (which can be required for some UX/features, see https://github.com/EddyVerbruggen/Custom-URL-scheme/issues/206#issuecomment-252182327)

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ You can find the library on `window.History`.
 `history` provides 3 different methods for creating a `history` object, depending on your environment.
 
 - `createBrowserHistory` is for use in modern web browsers that support the [HTML5 history API](http://diveintohtml5.info/history.html) (see [cross-browser compatibility](http://caniuse.com/#feat=history))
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -122,7 +122,8 @@ The `location` object implements a subset of [the `window.location` interface](h
 
 Locations may also have the following properties:
 
-- `location.state` - Some extra state for this location that does not reside in the URL (supported in `createBrowserHistory` and `createMemoryHistory`)
+- `location.state` - Some extra state for this location that does not reside in the URL (supported in `createBrowserHistory` and `
+`)
```
mjackson: Looks like this change wasn't intentional?

slorber: good catch

mjackson:
Thanks, @slorber!

PR #377, opened 2016-10-16
Added typings to use with typescript

mikaelwaltersson:
In line with the documentation and code for version 4.3.0.
The ones available in the Typings repository is only for version 2.


mjackson:
Thanks for the contribution, @mikaelwaltersson. I'm hesitant to merge this PR however, because I don't actually use TypeScript. Is there any way you can maintain these definitions in a separate library? Or do they need to be here?

cajoy:
Can we merge this changes? It prevents me from using this library because of missing typings :(

mjackson:
ping @mikaelwaltersson

mikaelwaltersson:
When packages lacks typings one needs to use a typings.json containing external sources for package typings and keep it up to date with package.json which can become quite messy. 

Bundled typings in packages make life much easier for TypeScript developers. But I get that it adds extra maintenance keeping the index.d.ts up to date when you update your library api. I will just keep using the fork if it's not merged, not a big deal for me. @cajoy could do the same.

mjackson:
> I will just keep using the fork if it's not merged, not a big deal for me. @cajoy could do the same.

Thanks, @mikaelwaltersson. That would definitely help me in the short term. In the long term I think we'll probably end up merging, but I'd like to see some more people express interest first.

Thanks :)

cajoy:
@mjackson  I believe you can publish Definitions into public repo so it is going to be available for everyone without fork: 

http://definitelytyped.org/guides/contributing.html

also take a look into https://blogs.msdn.microsoft.com/typescript/2016/06/15/the-future-of-declaration-files/

pqr:
> but I'd like to see some more people express interest first.
I do express my interest in it!

mjackson:
Thanks for the links, @cajoy. I definitely like the sound of having these typings in a `@types/history` package whose version numbers shadow the `history` package.

Any idea how this effort is coming along in the TypeScript community? Seems like the authors of the various types repos are already on board.

mjackson:
I'd prefer to keep this in @types for now. Please submit a PR over there :)

donaldpipowitch:
@mjackson I'd ask if this could be merged and would like to make a PR for `react-router`, too, if this is possible. @types is a nice project in theory and often very helpful, but it can be very troublesome. PRs can take a long time, because often there is someone missing who actually knows the lib which was typed behind it in-depth, so there is a certain level of uncertainty, if a change in the API is actually correct.

> I definitely like the sound of having these typings in a @types/history package whose version numbers shadow the history package.

That is probably the most troublesome problem, because from my point of view contributors have little control over this. I'm not even sure if we can create typings for prereleases which is needed for `react-router` v4 currently.


Different question: Would you consider a PR switching the code base to TypeScript completely? We currently plan to do the same for glamor. In my experience the tooling with TypeScript is superb for lib authors.

opavader:
@mjackson Please merge this PR.

mjackson:
> because often there is someone missing who actually knows the lib which was typed behind it in-depth

That's exactly what I'm afraid will happen here, except with a twist: no active contributors to either this repo or react-router use TypeScript, so likely nobody will know/care to maintain your TypeScript definitions when the underlying code changes.

As a developer, I'm surprised the TypeScript community doesn't already have a straight-forward way of dealing with this problem. Going around bugging every JS library author to merge code just to address the needs of a small subset of users doesn't seem like a great approach.

opavader:
My apologies if the request annoyed you. I like others here thought that keeping the type definition within the same repo was the standard & the best approach as it makes the versioning & installation extremely straight-forward while not affecting anything else at all (as done by all the newer libraries including redux, saga, immutable etc.). Typescript might be small community in react circle but thats how things scale, not long ago people had to ask to add npm config & publish while majority were using bower. 

As for maintenance, type definition is not that hard to update and isn't even needed if there are no breaking changes in the public api. So if required one of us will do the update as well. Alternative we can all use @mikaelwaltersson fork & ask him to keep the rest of code in sync with this repo.

Thanks.

donaldpipowitch:
> As for maintenance, type definition is not that hard to update and isn't even needed if there are no breaking changes in the public api.

That is my experience, too.

> no active contributors to either this repo or react-router use TypeScript, so likely nobody will know/care to maintain your TypeScript definitions when the underlying code changes.

We could change that, if we port the repositories to TypeScript :D Joke aside. The declaration format is quite easy and IMHO it is just a machine (and human) readable documentation. If you care about your documentation and maintain it, it should be quite easy to maintain the declaration files, too.

johannordin:
+1 for including it in repo. its a good documentation. the current declaration on @types is only 90 loc, so maintanance should be rather easy. 
anyway, thanks for a good repo!

donaldpipowitch:
> its a good documentation.

Just to illustrate this: some hours ago typing declarations for `react-router-dom` v4 were published. Here we can see the interface for [`match`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/react-router/index.d.ts#L85). It has `url` and `path` declared as optional fields. This was unexpected for me and I searched for this interface in the [official docs](https://reacttraining.com/react-router/web/api/match) which are _really good_. _But_... I can't find something about if `url` and `path` can be `undefined`. I _guess_ not and I _think_ the typings are wrong, but my point is that with real typings we have _one way_ to declare something as optional or not and this is human- and machine-readable. I currently have to _read_ the docs to see if something is optional or not and it can be expressed in slightly different ways, because it is written in prose.

mjackson:
@donaldpipowitch I think you may be misunderstanding my core concern which no one has addressed. It is: I'm afraid that we'll end up with outdated typings in our library because nobody here knows/cares about how to use them. That's it.

If you're offering to step up and become a contributor and maintain them, great! But you need to keep on top of API changes and update them accordingly.

donaldpipowitch:
> If you're offering to step up and become a contributor and maintain them, great! But you need to keep on top of API changes and update them accordingly.

Why not, we can try that :) We started with adding declarations in Glamor in the same way.

If it could be possible for other maintainers to ping me when they knowingly change the API (or before a public release), I hope it would be not to hard to keep on top of API changes. As far as I can see the API is quite stable currently, right?

PR #381, opened 2016-10-28
Use value-equal instead of deepEqual

### LINKED ISSUE #378: deepEqual is broken when A is object and B is null
I'm running into this an issue w/ react-router v4 and I tracked it down here.

The problem is here, I believe: https://github.com/mjackson/history/blob/929d5dca576512ab698486c98a3cb063e9a17ca9/modules/LocationUtils.js#L64-L66

```
     TypeError: Can't call method on  null
       exports (http://localhost:4002/js/vendor.bundle.js:1043)
       exports (http://localhost:4002/js/vendor.bundle.js:799)
       keys (http://localhost:4002/js/vendor.bundle.js:25618)
       looseEqual (http://localhost:4002/js/bundle.js:744)
       locationsAreEqual (http://localhost:4002/js/bundle.js:757)
# StaticRouter's componentWillReceiveProps
       componentWillReceiveProps (http://localhost:4002/js/bundle.js:102)
```

The next (or B) state property is null, but because A's state is an object, deepEqual attempts to call `Object.keys(null)`

```
state {"pathname":"/signin","search":"","hash":"","state":{"from":{"pathname":"/home","search":"","hash":"","state":null,"query":null}},"query":null}
next {"pathname":"/signin","search":"","hash":"","state":null,"query":null}
```

I don't know if I'm doing something wrong that's causing `state` to be null, but `deepEqual` should probably be resilient to this. Thoughts?

aaronjensen:
Fixes #378

### DIFF
--- a/modules/LocationUtils.js
+++ b/modules/LocationUtils.js
@@ -1,4 +1,5 @@
 import resolvePathname from 'resolve-pathname'
+import valueEqual from 'value-equal'
 import { parsePath } from './PathUtils'
 
 export const createLocation = (path, state, key, currentLocation) => {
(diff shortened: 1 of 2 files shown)


aaronjensen:
By the way, why leave `node_modules` out of `.gitignore` in your projects? It adds an annoying step for me to contribute.

aaronjensen:
@mjackson here you go. Aside from maybe size is there an advantage to `valueEqual` over `lodash.isequal`?

mjackson:
Thanks, @aaronjensen! Looks good.

> By the way, why leave node_modules out of .gitignore in your projects? It adds an annoying step for me to contribute.

I configured my [global `.gitignore`](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore) a long time ago and never looked back. This way I find I only have stuff in `.gitignore` that is relevant to that project and not to every project I ever work on.

mjackson:
> Aside from maybe size is there an advantage to valueEqual over lodash.isequal?

Size and simplicity are the main benefits. I don't need all of the guarantees `lodash.isequal` makes. Just a simple "are these two values equal?" is good enough in most cases.

aaronjensen:
> I configured my global .gitignore a long time ago and never looked back. This way I find I only have stuff in .gitignore that is relevant to that project and not to every project I ever work on.

Gotcha, I figured that may be what you did. I still adhere to the philosophy that a project's `.gitignore` are for files that it generates naturally and my global is for stuff my OS or other personal apps generate. I can see the argument either way.

aaronjensen:
@mjackson could you release a new `history` version when you get a chance please so we can start using `state`? Thanks!

mjackson:
Done! Released in version 4.4.0

PR #403, opened 2017-01-05
Add pop() to memory history 

LeoLeBras:
Hi,

First of all, thanks for your amazing work! :rocket:

I started using `history` with [`<NavigationExperimental />`](https://github.com/facebook/react-native/tree/9ee815f6b52e0c2417c04e5a05e1e31df26daed2/Libraries/NavigationExperimental) in react-native. `NavigationExperimental` provides two different methods to go back to a previous route: [`jumpTo()`](https://github.com/facebook/react-native/blob/9ee815f6b52e0c2417c04e5a05e1e31df26daed2/Libraries/NavigationExperimental/NavigationStateUtils.js#L110-L116) (similar to `go()` with `createMemoryHistory()` ) and [`pop()`](https://github.com/facebook/react-native/blob/9ee815f6b52e0c2417c04e5a05e1e31df26daed2/Libraries/NavigationExperimental/NavigationStateUtils.js#L76-L92). However: this one is missing in history. The goal of `pop()` is to update the index and pop out a route from the entries (when `go()` just updates the index in history).

This PR is a simple implementation of `pop()` method.

Thanks


mjackson:
AFAICT this does the same thing as `history.goBack()` or `history.go(-1)`, is that right?

mjackson:
I believe this does the same thing as API that we already have. Closing.

LeoLeBras:
Hi! Sorry for the late reply! 

As you said, `pop()` and `history.goBack()` or `history.go(-1)` act the same way. The only difference I tried to highlight is how they act under the hood.

In react-native for instance, we can use [`navigator.pop()`](https://facebook.github.io/react-native/docs/navigator.html#popn) to go back and then unmount the current scene. `navigator.jumpBack()` does the same thing as `navigator.pop()` but it doesn’t unmount the current scene while going back. This mean that we can then come back to this scene afterwards through a “next” button for instance. With `navigator.pop()`, after you’ve triggered the back action, you can’t come back again on the scene you just left.

Therefore, while we do have `navigator.jumpBack()` in `react-native` that acts like `goBack()` in history, we do not have a `pop()` method in history that could suit the need of unmounting the scene (like `navigator.pop()` does).

The goal of this PR is to add this `pop()` method so that `history` could act like `navigator.pop()` if we use it in `react-native`. For instance you’ll have: 

```js
const initial = {
  ...history,
  index: 1,
  entries: [{ pathanme: '/scene1' }, { pathname: '/scene2' }],
}

const afterGoBack = {
  ...history,
  index: 0,
  entries: [{ pathanme: '/scene1' }, { pathname: '/scene2' }],
}

const afterPop = {
  ...history,
  index: 0,
  entries: [{ pathanme: '/scene1' }],
}
```

That's why we do need these two types of “back methods” in `react-native`, and it would be great that history supports them.

LeoLeBras:
@mjackson ?

modules/__tests__/TestSequences/Pop.js
```diff
@@ -0,0 +1,33 @@
...
+      history.pop()
+    },
+    (location, action) => {
+      expect(action).toEqual('POP')
+      expect(location).toMatch({
+        pathname: '/'
+      })
+      expect(history.entries).toMatch([{
+        pathname: '/',
+      }])
```
mjackson: Why is this last assertion important? Is your app reading from `history.entries`? Why can't your app do what it needs to do at this point with just the current `history.location`?

mjackson:
Hey, sorry for the slowness. Been super busy :) I asked a question on the PR.

modules/__tests__/TestSequences/Pop.js
LeoLeBras: Yes, my app uses `history.entries` and can’t do what it needs with just
the current `history.location`. Let me explain:

Among the different navigation components available in React Native we have
a navigation a state` :

```js
// Build navigation state
const navigationState = {
  index: 0,
  routes: [
    { key: 'page_1' },
    { key: 'page_2' },
  ],
}
// Build the navigation component with NavigationExperimental API
const Navigation = (props) =>  (
  <NavigationCardStack
    {...props}
    navigationState={navigationState}
  />
)
```

or:

```js
// Build navigation state
const routes = [
  { key: 'page_1' },
  { key: 'page_2' },
]
const navigationState = {
  initialRoute: routes[0]
  initialRouteStack: routes,
}
// Build the navigation component with <Navigator />
const Navigation = (props) => (
  <Navigator
    {...props}
    {...navigationState}
  />
)
```

In order to work with `history`, `NavigationExperimental` and `Navigator` need
to depend on `history.entries` properties.

Indeed, `react-native` offers us two types of “back methods”. The first one allows
us to transition back and unmount the current scene. It means that we can't then
come back to this scene. The other one allow us to  jump backward without unmounting
the current scene and later jump forward to this scene. 

If we come back to my example,  in order to check these differences,
the current `history.location` is not enough, I need to analyze `history.entries`.

Here is a more detailed example:

First at all, we build the initial state.
```js
const initialHistory = {
  ...history,
  index: 1,
  entries: [{ pathanme: '/scene1' }, { pathname: '/scene2' }],
}
const initialNavigationState = {
  index: 1,
  routes: [{ key: '/scene1' }, { scene: '/scene2' }]
}
```

**Case 1.** After the call of `history.pop()`, the navigator transitions back
and unmounts the current scene. After that, we can't go back to `/scene2`
with `history.goForward()` because the entry was deleted in history and
the scene was unmounted in navigator.
```js
const historyAfterPop = {
  index: 0,
  routes: [{ key: '/scene1' }],
}
const navigationStateAfterPop = {
  index: 0,
  routes: [{ key: '/scene1' }],
}
```

**Case 2** After the call of `history.goBack()`, the navigator jump backward
without unmounting the current scene. In this case, we can jump forward
to `/scene1` later. `/scene2` is always mounted in `react-native`’s navigator.
```js
const historyAfterGoBack = {
  index: 0,
  routes: [{ key: '/scene1' }, { key: '/scene2' }],
}
const navigationStateGoBack = {
  index: 0,
  routes: [{ key: '/scene1' }, { key: '/scene2' }],
}
```

I hope you will find this understandable. It would be incredible to be able to use `history`
with `react-native`! :tada:.

LeoLeBras: @mjackson ? 😄

mjackson:
Hey @LeoLeBras :) We're working on getting great support for React Native in React Router v4. We've already got code that works in [our `react-router-native` package](https://github.com/ReactTraining/react-router/tree/v4/packages/react-router-native) which we'll be releasing this next week.

Would you be able to take a look and test it out? I think there may be something here, but I'm still not 100% sure.

mjackson:
Closing due to inactivity.

PR #405, opened 2017-01-10
Ensure that the provided `basename` starts with a forward slash `/`

mindjuice:
I was going to add a unit test for this, but noticed that adding proper unit tests for `basename` would be a lot more work than just adding one new test for this change, so this PR does not include new tests.  All existing tests still pass though.

I also had to break out `basename` from the destructuring assignment, which isn't as nice looking.  Let me know if you prefer this done a different way though.


mjackson:
👍 Thanks, @mindjuice

PR #413, opened 2017-01-16
Fix basename being case sensitive

fhelwanger:
I'm using `react-router` v3 with SSR, using `match` both on client and server.

When I use `match` on client, it tries to create location using `history.getCurrentLocation`: https://github.com/ReactTraining/react-router/blob/master/modules/match.js#L34

In this scenario, the basename is case sensitive. If it's not typed exactly by the user, the url won't be split between `pathname` and `basename`.

Here's an example: let's say I configured basename to be `/MyBaseName`. If the user types `http://myserver.com/mybasename` it won't match. It's the same if I configure basename all lowercase and the user type any char uppercase. Also, it's not fair to ask a mobile user to type each character with the exact case, they simple won't.

This is the only scenario where I found routes to be case sensitive, so I think it's a bug.

Please help me if my thoughts are wrong 😄

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,7 @@
+## [HEAD]
+
+- **Bugfix:** Fix basename being case sensitive ([#413])
+
 ## [v3.2.0]
 > Sep 1, 2016
 
--- a/modules/useBasename.js
+++ b/modules/useBasename.js
@@ -11,7 +11,7 @@ const useBasename = (createHistory) =>
         return location
 
       if (basename && location.basename == null) {
-        if (location.pathname.indexOf(basename) === 0) {
+        if (location.pathname.toLowerCase().indexOf(basename.toLowerCase()) === 0) {
           location.pathname = location.pathname.substring(basename.length)
           location.basename = basename
 


mjackson:
`basename` is designed to be a string-for-string replacement of the URL, but I don't see any reason it needs to be case-sensitive. I'm not really keen on maintaining a 3.x branch though. 4.x has been out for a long time now.

fhelwanger:
Yeah, I see your point, but react-router v3 is the stable version right now and it depends on history v3.

Please correct me if I'm wrong, but I don't think changing react-router v3 to use history v4 is viable at this point. Therefore, at least while react-router v4 isn't out of beta, I think it's reasonable to keep a maintenance branch for history v3.

If there's no reason to basename being case sensitive, I'd be very grateful if this could be merged.

mjackson:
Sure, will release soon.

mjackson:
Published in 3.3.0. Thank you!

fhelwanger:
Thank you very much @mjackson!

I'm very grateful for all your hard work on OSS :smile:

Issue #367, opened 2016-09-14
Blocked hash history changes from a href are still changed in URL bar


grassick:
I block a navigation using block and it correctly does not call the location change listener. 

However, the URL in the URL bar is in fact changed which is really confusing to the user and might lead to bookmarking the wrong page. I'm using hash history.

mjackson:
We've got a tests for this:

- [block a `PUSH`](https://github.com/mjackson/history/blob/8450e07b00ef08f10aba2fd6bd94d337e9cc07e6/modules/__tests__/HashHistory-test.js#L89-L93)
- [block a `POP`](https://github.com/mjackson/history/blob/8450e07b00ef08f10aba2fd6bd94d337e9cc07e6/modules/__tests__/HashHistory-test.js#L95-L99)

both of which are passing. Can you please provide a case that fails?

grassick:
My case is in Chrome when I click on a `<a href="#/somelinkhere">link</a>`. I'm not doing a push, but am clicking directly on an href link. The transition is correctly blocked in every respect except for the URL bar of the browser where it still changes.

ryanflorence:
does it go back to the correct URL after you cancel the prompt?

grassick:
Nope! It stays at the one that was blocked.

dlvx:
Same issue here, happens when using getUserConfirmation as well, seems in this case the revertPop(location) is not working but I could be wrong

socketwiz:
I'm seeing the same issue as well when trying to use with react-router and using Link to navigate.

```
        let history = useRouterHistory(() => createHashHistory({
            getUserConfirmation(message, callback) {
                return callback(confirm(message));
            }
        }))();
        const unblock = history.block('Are you sure you want to leave this page?');
```

It does present the confirmation message, but its already navigated to the destination before this occurs.  Pressing OK or Cancel just seems to dismiss the confirmation dialog.  It does not navigate back to where you came from, which doesn't really matter because its way to late by this point, it should never have navigated at all.

I'm using hashHistory not browserHistory if that makes a difference.

mjackson:
Would really appreciate some help here if anyone has the time to pitch in. Sounds like there's a legitimate issue. A failing test case here would be invaluable.

socketwiz:
[...]

PR #422, opened 2017-02-02
Add failing test for #367

grassick:
Here is the failing test that you asked for. It fails to restore the hash of the URL bar correctly when the change is blocked (at least in Chrome)


mjackson:
Thanks, @grassick. Unfortunately, I don't have time to maintain v3 any more. So if you'd like to get this merged you'll have to submit the fix in this PR as well.

PR #432, opened 2017-02-16
Remove trailing slash from basename

pshrmn:
If the basename passed to a history has a trailing slash, it is removed.

```js
'' => ''
'/' => ''
'/nested/site' => '/nested/site'
'/nested/site/' => '/nested/site'
```

I think that this is the best solution. If basenames with a trailing slash are left as they are provided, then that complicates both a) joining paths and b) stripping the prefix when getting the location from `window.location`. By removing the trailing slash, path joining is simple and you don't have to worry about creating a "relative" location in `getDOMLocation`.


mjackson:
Thanks, @pshrmn 👍

PR #433, opened 2017-02-16
Convert slash basename to empty string

pshrmn:
As I mentioned here https://github.com/ReactTraining/history/commit/d91c52b6609c6c30e77f74a195816b16bf4aabac#commitcomment-20916643, I should have tested this before.


mjackson:
Looks great, thank you 👍

PR #436, opened 2017-02-22
Fix multiple unlistens

aij:
Log a warning instead of decrementing listenerCount again
when the same unlisten function is called multiple times.

Also avoid adding event listeners a second time when
listenerCount is decreased to 1.

Also fix tests to avoid calling unlisten twice.

### DIFF
--- a/modules/__tests__/TestSequences/execSteps.js
+++ b/modules/__tests__/TestSequences/execSteps.js
@@ -3,6 +3,8 @@ const execSteps = (steps, history, done) => {
 
   const cleanup = (...args) => {
     unlisten()
+    // Avoid unsubscribing a second time when we double cleanup.
+    unlisten = () => {}
     done(...args)
   }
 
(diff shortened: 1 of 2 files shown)


mjackson:
I've never had an issue with `unlisten` being called twice. And if there is an issue, I'd prefer a PR fixing the issue instead of making `unlisten` a no-op.

aij:
The issue I was dealing with turned out to be something else, but it would have been nice to know that it wasn't caused by some other piece of code *somewhere* accidentally calling unlisten twice.

And actually, the tests are calling unlisten twice, which is why a new history object needs to be created for each one...

aij:
Would a PR that makes it a warning without changing the behavior be preferable/acceptable?

mjackson:
Sure, @aij. If you insert a warning about calling `unlisten` twice, and we start seeing that error all over the place in our tests, then I think we should fix it :)

PR #439, opened 2017-02-26
Fix Security error during accessing session storage when cookies are disabled

okovpashko:


mjackson:
Thanks, @okovpashko!

salzhrani:
is it possible to release this?

eldh:
@salzhrani This is released as v3.3.0, afaict.

salzhrani:
I think it is not
https://unpkg.com/history@3.3.0/lib/DOMStateStorage.js

eldh:
@salzhrani Ah, you're right! I just just looked at release date... In that case +1. Would be much appreciated. :)

salzhrani:
@mjackson friendly reminder

InsaneViku:
Hi @mjackson, thanks for merging this (and thanks @okovpashko for the patch)! Consider this another friendly reminder to push another v3 release for this fix. Here's an issue that opened up as well to request this: https://github.com/ReactTraining/history/issues/462. This would help immensely!

okovpashko:
@mjackson please elaborate and release this fix. 
Recently we ended up creating the same bug in our issue tracker again. My colleague started fixing it and stumbled upon this PR.
Even I forgot that I fixed it about a year and a half ago.

PR #440, opened 2017-02-28
Make unlisten function idempotent

aij:
Instead of continuing to decrement listenerCount each
time it is called.


aij:
Ugh, tests didn't even run there. `yarn` failed to install.

Tests are passing when I run them locally FWIW.

mjackson:

PR #444, opened 2017-03-09
Failing (and succeding) tests for relative urls

aij:
Add some tests for relative URLs, to show what works and what doesn't.

### DIFF
--- a/modules/__tests__/LocationUtils-test.js
+++ b/modules/__tests__/LocationUtils-test.js
@@ -199,35 +199,5 @@ describe('createLocation', () => {
         })
       })
     })
-    describe('can resolve a relative fragment when there is no query', () => {
-      const currentNoQuery = {
-        pathname: '/The/Old/Path',
-        search: '',
-        hash: '#fragment'
-      };
-
-      describe('given as a string', () => {
-        it('has the correct properties', () => {
-          expect(
-            createLocation('#foo', undefined, undefined, currentNoQuery)
-          ).toMatch({
-            pathname: '/The/Old/Path',
-            search: '',
-            hash: '#foo'
-          })
-        })
-      })
-      describe('given as an object', () => {
-        it('has the correct properties', () => {
-          expect(
-            createLocation({ hash: '#foo' }, undefined, undefined, currentNoQuery)
-          ).toMatch({
-            pathname: '/The/Old/Path',
-            search: '',
-            hash: '#foo'
-          })
-        })
-      })
-    })
   })
 })
--- a/modules/__tests__/LocationUtils-test.js
+++ b/modules/__tests__/LocationUtils-test.js
@@ -119,21 +119,10 @@ describe('createLocation', () => {
       hash: '#fragment'
     };
 
-    const createRelative = url => createLocation(url, undefined, undefined, currentLocation)
-
     describe('can resolve a relative pathname', () => {
-      describe('given as a string', () => {
+      describe('without any dots', () => {
         it('has the correct properties', () => {
-          expect(createRelative('new/path')).toMatch({
-            pathname: '/The/Old/new/path',
-            search: '',
-            hash: ''
-          })
-        })
-      })
-      describe('given as an object', () => {
-        it('has the correct properties', () => {
-          expect(createRelative({ pathname: 'new/path' })).toMatch({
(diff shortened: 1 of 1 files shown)


mjackson:
Thanks, @aij. A good set of tests is invaluable.

modules/__tests__/LocationUtils-test.js
```diff
@@ -111,4 +111,123 @@ describe('createLocation', () => {
...
+        it('has the correct properties', () => {
+          expect(createRelative({ hash: '#foo' })).toMatch({
+            pathname: '/The/Old/Path',
+            search: '?theQuery',
+            hash: '#foo'
+          })
+        })
+      })
+    })
+    describe('can resolve a relative fragment when there is no query', () => {
```
mjackson: `describe('when there is no query')`

modules/__tests__/LocationUtils-test.js
```diff
@@ -111,4 +111,123 @@ describe('createLocation', () => {
...
+      const currentNoQuery = {
+        pathname: '/The/Old/Path',
+        search: '',
+        hash: '#fragment'
+      };
+
+      describe('given as a string', () => {
+        it('has the correct properties', () => {
```
mjackson: `it('resolves a relative fragment')`

modules/__tests__/LocationUtils-test.js
```diff
@@ -111,4 +111,123 @@ describe('createLocation', () => {
...
   })
+
+  describe('with a current location', () => {
+    const currentLocation = {
+      pathname: '/The/Old/Path',
+      search: '?theQuery',
+      hash: '#fragment'
+    };
+
+    const createRelative = url => createLocation(url, undefined, undefined, currentLocation)
```
mjackson: Let's use `createLocation` directly for consistency with the other tests in this file.

modules/__tests__/LocationUtils-test.js
```diff
@@ -111,4 +111,123 @@ describe('createLocation', () => {
...
+          expect(createRelative('new/path')).toMatch({
+            pathname: '/The/Old/new/path',
+            search: '',
+            hash: ''
+          })
+        })
+      })
+      describe('given as an object', () => {
+        it('has the correct properties', () => {
+          expect(createRelative({ pathname: 'new/path' })).toMatch({
```
mjackson: Having 2 tests that are identical with different forms of the argument feels tedious. Can we just test behavior once?

modules/__tests__/LocationUtils-test.js
aij: Ok, should currentLocation be inlined too, or should I leave it as a `const` to avoid repeating it in every test?

modules/__tests__/LocationUtils-test.js
aij: I agree. The only reason I did it was to match the other tests in this file. Should I leave the existing tests testing both forms, but change the new ones to test only the string form?

modules/__tests__/LocationUtils-test.js
aij: Note that I added these very last set of tests as a separate commit, "of dubious merit" just to show that it does work in the special case when the query is empty. Do you think these will be worth keeping (and renaming), or would we just drop them once the test above is passing?

And sorry for the delay in responding. I've been working on other things for a while.

modules/__tests__/LocationUtils-test.js
mjackson: Ah, good call to follow existing style. I'd like to slim down if we can and avoid duplication. You can stick with just the more convenient string form in these tests.

modules/__tests__/LocationUtils-test.js
mjackson: It's ok to leave `currentLocation` up top. I just like to have some consistency between different tests.

modules/__tests__/LocationUtils-test.js
mjackson: You can feel free to drop them if you don't think they're necessary. No worries about the slow response :)

taurose:
How about empty string `createLocation('', ...)`? It seems it should keep the pathname and query but drop the fragment, but currently the pathname defaults to `/` for all falsy values.

mjackson:

PR #445, opened 2017-03-11
Only encode/decode the pathname

pshrmn:
The `pathname` is the only part of the URL that React Router cares about, so that is the only part that I am confident in encoding/decoding.

A decoded `search` string that contains a `#` (or any of the other reserved characters) may cause bugs if we only encode it using `encodeURI`.

```js
// starting with the location
{
  pathname: '/',
  search: '?hashtag=#yolo',
  hash: ''
}
// this will create the encoded url /?hashtag=#yolo
// which will be parsed to create the location object
{
  pathname: '/',
  search: '?hashtag=',
  hash: '#yolo'
}
```

If we expect the user to be producing an encoded search string (which the different query modules will automatically do), then we can expect them to produce a valid `search` string.

### DIFF
--- a/modules/PathUtils.js
+++ b/modules/PathUtils.js
@@ -15,7 +15,6 @@ export const parsePath = (path) => {
   let search = ''
   let hash = ''
 
-  pathname = decodeURI(pathname)
   const hashIndex = pathname.indexOf('#')
   if (hashIndex !== -1) {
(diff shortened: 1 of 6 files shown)


mjackson:
Sounds good to me. Thanks for taking another look, @pshrmn

Issue #452, opened 2017-03-21
Warn when pathname does not contain basename


pshrmn:
Given a history with a basename:

```js
const history = createBrowserHistory({ basename: '/base' })
```

When the history gets the location from the URL, it will strip the basename off of the pathname.

```js
if (basename)
  path = stripPrefix(path, basename)
```

This means that a pathname without the basename is treated the same as a pathname with the basename.

```js
"/base/my-page" == "/my-page"
```

It could be argued that this problem exists only when a user is improperly creating the `history` (you should only create a history with a basename when you are on a page with that basename), but at the very least I think that a warning would be helpful.

```js
warning(
  !(basename && pathname.indexOf(basename) !== 0),
  "You are attempting to use a basename on a page whose URL has no basename."
)
```

mjackson:
Sure, I think that makes sense to warn in this case.

PR #453, opened 2017-03-22
Warn when path does not contain basename

pshrmn:
As brought up in #452 

I don't love the tests I wrote, but I added them because I saw that there were test sequences for other warnings. Unfortunately, the only great way to catch the warning was during the creation of a history object, which I couldn't do in a test sequence (because that expects the history to already exist).

I also added some `pushState` calls in the `createHref` tests so that the test output doesn't get littered with warnings.

### DIFF
--- a/modules/createBrowserHistory.js
+++ b/modules/createBrowserHistory.js
@@ -59,6 +59,12 @@ const createBrowserHistory = (props = {}) => {
 
     let path = pathname + search + hash
 
+    warning(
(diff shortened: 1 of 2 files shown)


mjackson:
Thanks for the PR, @pshrmn! Honestly, I'd be fine merging this w/out the tests if you'd like to remove them.

pshrmn:
Just rebased those out.

mjackson:
🙏

PR #455, opened 2017-03-31
Disable creating `href` for memory history

tricknotes:
When using memory history, the href that created by history is unnecessary for users.
If users knows the URL from href, they will visit and meet route not found :<
The routing structure should be hidden in internal.


mjackson:
We use URLs from memory history to represent the state of the app on native and in other environments. The fact that your users are visiting your memory history URLs is not a limitation of this library.

PR #456, opened 2017-04-01
avoid adding leading slash to already full qualified urls to origin.

joeuy:
This change avoids adding unnecessary leading slashes when basename is set to window.location.origin.

There was some discussion about why this is useful here: https://github.com/ReactTraining/react-router/issues/4885#issuecomment-290818502

** And here is an example that further ilustrates why this is useful:**
http://jsbin.com/deyovujamu/edit?html,js,output

Things to note in the example:
- all image links are relative
- we specify the cdn host name using `<base href="https://c1.staticflickr.com">`
- Images are loading correctly !!!
- links are broken. instead of referring to `http://null.jsbin.com/about`, it will point to `https://c1.staticflickr.com/http://null.jsbin.com/about` that's because history always add leading slash, and then the browser adds the base href.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,7 @@ The options that each `create` method takes, along with its default values, are:
 ```js
 createBrowserHistory({
   basename: '',             // The base URL of the app (see below)
+  includeOrigin: false,     // Set true to include the origin in generated urls. 
   forceRefresh: false,      // Set true to force full page refreshes
   keyLength: 6,             // The length of location.key
   // A function to use to confirm navigation with the user (see below)
--- a/modules/DOMUtils.js
+++ b/modules/DOMUtils.js
@@ -36,6 +36,14 @@ export const supportsHistory = () => {
 }
 
 /**
+ * Polyfill returns window.location.origin 
+ * https://developer.mozilla.org/en-US/docs/Web/API/Window/location#Browser_compatibility
+ */
+export const getOrigin = () =>
+  window.location.origin ||
+  window.location.protocol + "//" + window.location.hostname + (window.location.port ? ':' + window.location.port : '');
+
+/**
  * Returns true if browser fires popstate on hash change.
  * IE10 and IE11 do not.
  */
--- a/modules/__tests__/createHref-test.js
+++ b/modules/__tests__/createHref-test.js
@@ -1,4 +1,5 @@
 import expect from 'expect'
+import { getOrigin } from '../DOMUtils'
 import createBrowserHistory from '../createBrowserHistory'
 import createHashHistory from '../createHashHistory'
 import createMemoryHistory from '../createMemoryHistory'
@@ -69,6 +70,34 @@ describe('a browser history', () => {
       })
 
       expect(href).toEqual('/the/path?the=query#the-hash')
+    })
+  })
+
+  describe('with includeOrigin', () => {
+    let history
+
+    it('knows how to create hrefs', () => {
+      history = createBrowserHistory({ includeOrigin: true })
+      const href = history.createHref({
+        pathname: '/the/path',
+        search: '?the=query',
+        hash: '#the-hash'
+      })
+
+      const origin = getOrigin();
+      expect(href).toEqual(`${origin}/the/path?the=query#the-hash`)
+    })
+
+    it('and basename, knows how to create hrefs', () => {
+      history = createBrowserHistory({ includeOrigin: true, basename: '/the/base' })
+      const href = history.createHref({
+        pathname: '/the/path',
+        search: '?the=query',
+        hash: '#the-hash'
+      })
+
+      const origin = getOrigin();
+      expect(href).toEqual(`${origin}/the/base/the/path?the=query#the-hash`)
     })
   })
 
--- a/modules/createBrowserHistory.js
+++ b/modules/createBrowserHistory.js
@@ -51,7 +51,9 @@ const createBrowserHistory = (props = {}) => {
     getUserConfirmation = getConfirmation,
     keyLength = 6
   } = props
-  const basename = props.basename ? stripTrailingSlash(addLeadingSlash(props.basename)) : ''
+  const rawBasename = props.basename || '';
+  const isFullUrl = rawBasename.indexOf(window.location.origin) === 0;
+  const basename = stripTrailingSlash(isFullUrl ? rawBasename : addLeadingSlash(rawBasename));
 
   const getDOMLocation = (historyState) => {
     const { key, state } = (historyState || {})
--- a/modules/createBrowserHistory.js
+++ b/modules/createBrowserHistory.js
@@ -15,6 +15,7 @@ import {
   removeEventListener,
   getConfirmation,
   supportsHistory,
+  getOrigin,
   supportsPopStateOnHashChange,
   isExtraneousPopstateEvent
 } from './DOMUtils'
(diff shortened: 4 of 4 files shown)


mjackson:
The `basename` isn't supposed to include the origin at all, just the beginning portion of the `pathname`. I'm worried that you're trying to give too much meaning to `basename`. It's literally just a string that precedes the `pathname`, that's it.

joeuy:
The statement "It's literally just a string that precedes the pathname," holds valid under this change. 

The only difference here is that history  avoids adding a leading slash, if the prefix includes the current origin. This enables very powerful scenarios like the one described in the example: http://jsbin.com/deyovujamu/edit?html,js,output

pshrmn:
Previous discussion ReactTraining/react-router#4885

joeuy:
@mjackson it would be nice to know your thoughts. I think the example is a fair use case: 
- it enables nice CDN integration across environments for a production app. (see https://github.com/ReactTraining/react-router/issues/4885#issuecomment-290818502)
- it is not a breaking change.

Thanks.

scriby:
I would love to see this change make it in.

It's a simple change, reverse compatible with the existing code, and enables a new use case. 

You are free to question whether the new use case is "doing it wrong". It certainly isn't intended to be used by all consumers of this module; it just provides another option so users can decide what's appropriate for their situation.

Thanks for your consideration,

Chris

scriby:
I disagree that this usage of the base href is unintended.

https://html.spec.whatwg.org/multipage/semantics.html#the-base-element does indeed state that

> The base element allows authors to specify the document base URL for the purposes of **parsing URLs**, and the name of the default browsing context for the purposes of following hyperlinks. The element does not represent any content beyond this information.

An older version of the spec (https://www.w3.org/TR/html5/document-metadata.html#the-base-element) states:

> The base element allows authors to specify the document base URL for the purposes of **resolving relative URLs**, and the name of the default browsing context for the purposes of following hyperlinks. The element does not represent any content beyond this information.

Browsers still clearly use the base href to resolve relative URLs; and changing this today would be a huge breaking change. I don't know why the language was changed to read "parsing URLs" instead of "resolving relative URLs", but looking at https://html.spec.whatwg.org/multipage/infrastructure.html#parse-a-url I'd argue that resolving relative URLs is part of what they mean by parsing a URL.

Thanks,

Chris

pshrmn:
It is _so_ simple to just implement a function that prepends your cnd url onto a path that it just doesn't seem worth the hassle to break navigation for.

```js
// staticURL.js
const CDN = 'https://my.cdn.example.com'
export default function staticURL(path) {
  return CND + path
}

// img.js
import staticURL from './staticURL'

const Thing = () => (
  <img src={staticURL('/12312321.png')} />
)
```

Also, this PR is missing tests that show this, but it would not work with the current `getDOMLocation` implementation.

scriby:
@pshrmm also consider that there are assets referenced by CSS that are not dynamic at runtime.

pshrmn:
Do you mean something like this?

```css
.thing {
  background: url('my-background.png');
}
```

https://www.w3.org/TR/REC-CSS1/#url
> Partial URLs are interpreted relative to the source of the style sheet, not relative to the document

scriby:
What if you are using webpack and it injects the CSS as a style tag into the document?

pshrmn:
[`output.publicPath`](https://webpack.js.org/configuration/output/#output-publicpath)

joeuy:
@pshrmn lets be realistic, its not just about images, its everything you have to re-stamp: 
- scripts
- css
- images
- any other relative resource that goes to the cdn, think about background-image, etc

plus the CDN is different for testing, integration and production environment.... so you have to add yet more stuff to it. 

you are suggesting hundreds of lines of changes in a big project, vs a simple line change in the index.html file. 

>  it would not work with the current getDOMLocation implementation.
could you explain whats the problem? everything seems fine to me.

pshrmn:
`getDOMLocation` creates a path by concatentation `location.pathname + location.search + location.hash`. If there is a `basename` defined, then that should be stripped off of the `path`.

When `basename` includes the origin, this will never happen and the `path` will be the concatenated properties. Now, if the basename is _only_ the origin (`http://www.example.com`), then this does not matter. If the basename includes both the origin and an actual basename (`http:///www.example.com/base`), then this would fail to remove the basename (`/base/real/location` should be stripped to be `/real/location`).

---

A more reasonable request than this change would be to have `history` include an `origin` option. Then, `createHref` would be modified to prepend `window.origin` to its returned path.

I still think that using `<base>` to specify a CDN is a hack and the fact that it breaks navigation should be a giant red flag, but at least that approach wouldn't break anything.

scriby:
Yeah, you can totally use output.publicPath, but you have admit it's pretty cool to be able to set the base path on the fly. It lets you do cool stuff like use the same deployment package in test, stage, and prod. Decoupling the location of the assets from the asset package itself can't be a bad thing.

joeuy:
@pshrmn I modified the PR to be more specific, and only enable the scenario described here: 
- it will address your concerns about getDOMLocation 
- it will enable the powerful CDN scenario.
- it will further limit the impact of the change.

@mjackson looking forward to hearing your opinion. 

Thanks.

pshrmn:
Now you have to pick between the origin and having an actual basename.

My suggestion was that _if_ this is enabled, it should be done through a separate option. A basename should only be used to describe the part of the `pathname` that isn't used for routing.

```js
const regularHistory = createBrowserHistory()
const fullHistory = createBrowserHistory({ origin: true })
// when window.location.origin === 'http://www.example.com'
regularHistory.createHref('/test') // returns /test
fullHistory.createHref('/test') // returns http://www.example.com/test
```

Also, IE didn't support `window.location.origin` until 11, so that would need a polyfill.

joeuy:
@pshrmn thanks for you last comment. Updated PR with your feedback.

scriby:
I'll make one final argument for this. In essence we're talking about two approaches:

1. Use the base href as the base for navigation, and use absolute URLs for static assets.
2. Use the base href for static assets, and use absolute URLs for navigation.

I don't understand on what grounds it can be said that one of those approaches is legitimate and the other "a hack".

The argument for the second approach is that it's simpler to dynamically alter the URLs for navigation at runtime (via a simple router config) than it is to alter the URLs for static assets at runtime.

It's fine if it's not everyone's preferred approach, but it's nice if the router/history module support the option.

Thanks for your consideration,

Chris

joeuy:
@mjackson ping! any updates? thanks.

mjackson:
I agree with @pshrmn when he said

> A basename should only be used to describe the part of the pathname that isn't used for routing.

That's what I was trying to convey when I said "it's just the string that precedes the pathname".

`history` is a library that supports building single-page apps, on the *same* domain. All code in this library should be able to safely assume that. If we can't, we're going to run into all sorts of edge cases and probably not even be able to find them all until someone at some future date runs into them and shows up here complaining that everything's broken.

The good news for you is that since we're not doing anything at all with the origin you have total control over it. At runtime you can use a wrapper function to create URLs that point to your CDN. At build time you can use webpack's `publicPath` to point to the CDN. That's what it's for. From [the webpack docs](https://github.com/webpack/docs/wiki/configuration#outputpublicpath):

> The publicPath specifies the public URL address of the output files when referenced in a browser. For loaders that embed <script> or <link> tags or reference assets like images, publicPath is used as the href or url() to the file when it's different than their location on disk (as specified by path). **This can be helpful when you want to host some or all output files on a different domain or on a CDN**.

If you'd like to discuss a separate `origin` option, as @pshrmn [suggested](https://github.com/ReactTraining/history/pull/456#issuecomment-292336087) I'd be interested to hear that. But I'm not convinced that mixing origin and basename concepts is going to be productive.

PR #459, opened 2017-04-05
Smarter stripPrefix

pshrmn:
Per ReactTraining/react-router#4866

Does a case-insensitive match and will not strip the prefix if the prefix is a partial match (e.g., given the pathname `/prefix/pathname`, the basename `/pre` will not be stripped). Being more selective about not stripping partial basename matches is potentially breaking, but I would consider it more a bug that you could do that before.

@mjackson I believe that you have stated before that you don't like directly testing non-external APIs, but it felt like overkill to test these by triggering `getDOMLocation` calls when creating a browser/hash history.


mjackson:
Thanks for the PR, @pshrmn. This looks great. Thanks to you I think I'm finally getting some words to use to describe to people what a basename is. "The slash-delimited portion of the pathname that is not used for routing".

As for the tests, I think they'd be better if they used public API, yes. It does feel a little heavy, but tests that cover public API are the only ones that matter.

mjackson:
Idea: we can have 2 internal functions: `hasBasename(path, basename)` and `stripBasename(path, basename)`. Everywhere [we're using `path.indexOf(basename)`](https://github.com/ReactTraining/history/blob/fbf5f47f964a7df9ff69c99740667dbd763e3dd7/modules/createBrowserHistory.js#L63) we can use use `hasBasename` instead. And `stripBasename` can just use `hasBasename` internally to do a `path.substr` if needed.

pshrmn:
Alright, so there is a potential bug here that needs to be addressed before this is merged:

in `parsePath`, if the `path` is empty, then [`pathname` is set to `/`](https://github.com/ReactTraining/history/blob/master/modules/PathUtils.js#L14).
```js
let pathname = path || '/'
```
However, if the `path` has no `pathname`, but it includes a `search` or `hash` string, then `pathname` will end up being an empty string after those have been removed.

```js
path = '?test=ing'
pathname = '?test=ing'
// remove the search string
pathname = ''
```
When that happens, should `pathname` be set to `/`?

mjackson:
Yes, we should never have an empty pathname. The "empty" pathname is `/`.

modules/PathUtils.js
```diff
@@ -4,8 +4,11 @@ export const addLeadingSlash = (path) =>
 export const stripLeadingSlash = (path) =>
   path.charAt(0) === '/' ? path.substr(1) : path
 
-export const stripPrefix = (path, prefix) =>
-  path.indexOf(prefix) === 0 ? path.substr(prefix.length) : path
+export const hasBasename = (path, prefix) => 
+  (new RegExp('^' + prefix + '(\\/|\\?|#|$)', 'i')).test(path)
+
+export const stripBasename = (path, prefix) =>
+  hasBasename(path, prefix) ? path.substr(prefix.length) : path
```
mjackson: 👍

modules/__tests__/BrowserHistory-test.js
```diff
@@ -168,4 +169,42 @@ describeHistory('a browser history', () => {
       TestSequences.HashChangeTransitionHook(history, done)
     })
   })
+
+  describe('basename', () => {
```
mjackson: ❤️  these tests. Very easy to read.

modules/__tests__/BrowserHistory-test.js
```diff
@@ -168,4 +169,42 @@ describeHistory('a browser history', () => {
...
+    it('strips with no pathname, but with a search string', () => {
+      window.history.replaceState(null, null, '/prefix?a=b')
+      const history = createHistory({ basename: '/prefix' })
+      expect(history.location.pathname).toEqual('')
+    })
+
+    it('strips with no pathname, but with a hash string', () => {
+      window.history.replaceState(null, null, '/prefix#rest')
+      const history = createHistory({ basename: '/prefix' })
+      expect(history.location.pathname).toEqual('')
```
mjackson: `history.location.pathname` should be `/` here.

modules/createBrowserHistory.js
```diff
@@ -60,19 +60,15 @@ const createBrowserHistory = (props = {}) => {
...
     if (basename)
-      path = stripPrefix(path, basename)
+      path = stripBasename(path, basename)
 
-    return {
-      ...parsePath(path),
-      state,
-      key
-    }
+    return createLocation(path, state, key)
```
pshrmn: I couldn't just default `pathname` to `/` in `parsePath` when it was an empty string  because that would break the resolving that is done using `history.location` I'm not altogether satisfied with this approach because now the hash history location objects will have a `key` property. Maybe this isn't a big deal, but it is a side effect of the change.

mjackson: It's not a huge deal but I'd prefer it if hash history objects didn't have a `key` property. I think it would mislead people into thinking that we support key with hash history when we don't really.

It feels like this might be one of those changes that requires a more thorough refactoring. If that's the case, I'm happy to take it on. I've set aside some time today and tomorrow for OSS work.

pshrmn: I really hadn't thought about this before, but I just updated `createLocation` to only set the `key` property when the `key` value is truthy.

mjackson: 👍  That works!

modules/PathUtils.js
```diff
@@ -4,8 +4,11 @@ export const addLeadingSlash = (path) =>
 export const stripLeadingSlash = (path) =>
   path.charAt(0) === '/' ? path.substr(1) : path
 
-export const stripPrefix = (path, prefix) =>
-  path.indexOf(prefix) === 0 ? path.substr(prefix.length) : path
+export const hasBasename = (path, prefix) => 
+  (new RegExp('^' + prefix + '(\\/|\\?|#|$)', 'i')).test(path)
```
hmate9: This doesn't work if `prefix` has regex special symbols in them. Like `$`. It should be
```
export const hasBasename = (path, prefix) => path.startsWith(prefix)
```

pshrmn: A `basename` is meant to represent full segments, which `startsWith` cannot enforce on its own (the basename `/test` should not match the pathname `/testing`).

The `prefix` probably should be passed to a function that escapes any special regexp characters. I'm sure that a PR with tests would be appreciated for this.

hmate9: Thanks for the quick reply. I realised my mistake while makinging the fix. I created this PR for escaping special regex symbols: https://github.com/ReactTraining/history/pull/544

PR #465, opened 2017-04-21
No encode, decode in createLocation

pshrmn:
Removes the automatic encoding of `pathname`s, leaving that up to the browser.

Moves the decoding to `createLocation` because 1) that is now used by the `getDOMLocation` calls to create a location and 2) we should ensure that if a user pushes a location whose `pathname` is encoded, we decode it for consistency.


modules/__tests__/TestSequences/execSteps.js
```diff
@@ -21,6 +21,7 @@ const execSteps = (steps, history, done) => {
       if (index === steps.length)
         cleanup()
     } catch (error) {
+      console.log('caught error', error)
```
mjackson: Did you wanna leave this in?

pshrmn: Oh bother. That's the remnant of my debugging when I accidentally passed `done` to a `describe` call instead of an `it` call.

modules/__tests__/TestSequences/EncodedReservedCharacters.js
```diff
@@ -0,0 +1,46 @@
...
+        pathname: '/view/',
+        hash: '#abc'
+      })
+      // unencoded object
+      const pathname = '/view/#abc'
+      history.replace({ pathname })
+    },
+    (location) => {
+      expect(location).toMatch({
+        pathname: '/view/#abc'
```
pshrmn: I'm not actually sure about this test case. It passes, but I'm not sure if the behavior is desirable. This only happens if the user passes a bad location object.

modules/LocationUtils.js
```diff
@@ -33,7 +34,19 @@ export const createLocation = (path, state, key, currentLocation) => {
       location.state = state
   }
 
-  location.pathname = decodeURI(location.pathname)
+  try {
+    location.pathname = decodeURI(location.pathname)
+  } catch (e) {
+    if (e instanceof URIError) {
+      warning(
```
pshrmn: If we try to create a location with a bad percent-encoding, we swallow the error, keep `location.pathname` encoded, and log a warning.

modules/PathUtils.js
```diff
@@ -37,16 +37,32 @@ export const parsePath = (path) => {
...
 
+const validatePath = (path) => {
+  try {
+    decodeURI(path)
+  } catch(e) {
+    if (e instanceof URIError) {
+      throw new Error(
+        'You are attempting to create a path that cannot be decoded. ' +
+        'Please call encodeURI on the pathname.'
+      )
```
pshrmn: A slightly more useful error than the default `Uncaught URIError: URI malformed`

pshrmn:
Updated to just throw in `createLocation` when a `pathname` cannot be decoded. I believe that this makes the most sense to do, although it can be easily changed if you think otherwise @mjackson.

Internally, `createPath` is always passed a location object that was created by `createLocation`, so just having `createLocation` throw should be good enough. Having said that, `createPath` _is_ exported by `history`, so maybe it should throw on its own?

pshrmn:
The obvious issue with `history` throwing when it encounters a malformed URI is that developers will need to `try...catch` their code if they want to recover from bad percent-encoding. This would make it difficult to display an error page that looks like it is in their website.

```js
try {
  ReactDOM.render((
    <BrowserRouter>
      <App />
    </BrowserRouter
  ), holder)
} catch (e) {
  ReactDOM.render((
    <MalformedURIMessage />
  ), holder)
}
```

I still think that `history` should throw and we can point to the RFC 3986 [path parser rules](https://tools.ietf.org/html/rfc3986#section-3.3) to justify that, but there will still probably be people who want to use malformed URIs for some reason. I don't necessarily like the idea of automatically using the encoded `pathname` if the decode fails, but I think that it might be useful in the future to add an explicit "no decode" mode.

nil4:
> I still think that history should throw and we can point to the RFC 3986 path parser rules to justify that, but there will still probably be people who want to use malformed URIs for some reason

@pshrmn I think this is the right approach; an app using an invalid/malformed path can always be fixed to encode correctly. Thank you for fixing this.

modules/__tests__/TestSequences/EncodedReservedCharacters.js
mjackson: Seems like the kind of thing we don't actually need to test. i.e. if anyone actually tried to do this I'd probably just tell them to use `hash` instead of putting it on the `pathname`.

pshrmn:
Just bumping this since it has been a while. Whether or not this is the best approach, something definitely should be done here since the current implementation (made by myself :frowning_face:) breaks things.

nil4:
@pshrmn I would love to see this added. 👍

mjackson:
This looks good to me. Thanks, @pshrmn!

mjackson:
Seems like this should probably be a patch release since we're just fixing a bug we introduced in 4.6.1. Would you agree?

pshrmn:
I agree. More of a bug fix than "added functionality", so I don't think that it would warrant a minor bump.

pshrmn:
As a minor addendum, the next release would also include #459, but that is also pretty much just a bug fix, so I don't think that it warrants a minor bump.

pshrmn:
One more thing I just remembered is that #485 needs to be addressed before the next release.

timdorr:
OK, just got that one in for you.

@mjackson Did you want to push out a 4.6.2 release or add me to npm so I can take care of that for you? (you probably should add Ryan, BTW 😄)

mjackson:
Thanks, @timdorr. I'll go ahead and cut the release :)

PR #467, opened 2017-04-24
add option to include origin in generated urls.

joeuy:
Add a config option to include origin in generated urls.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,7 @@ The options that each `create` method takes, along with its default values, are:
 ```js
 createBrowserHistory({
   basename: '',             // The base URL of the app (see below)
+  includeOrigin: false,     // Set true to include the origin in generated urls. 
   forceRefresh: false,      // Set true to force full page refreshes
   keyLength: 6,             // The length of location.key
   // A function to use to confirm navigation with the user (see below)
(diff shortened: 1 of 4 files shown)


mjackson:
Why do you need this?

joeuy:
Hi, 
The justification for this change and discussion is coming from previous PR. 

> If you'd like to discuss a separate origin option, as @pshrmn suggested I'd be interested to hear that.
https://github.com/ReactTraining/history/pull/456#issuecomment-295810856

mjackson:
Why can't you just `document.origin + history.createHref('/more/stuff')`? It doesn't seem like we need to give you API for this.

joeuy:
The main reason I could think about is that browser history is in the business of creating urls, its better for clients to rely on history module which will abstract away any browser difference etc, and give you what you need. 

Also the main motivation for this is to use in react-router, and I was told to get the change done here first.

joeuy:
@mjackson any comment?

mjackson:
As I said in https://github.com/ReactTraining/history/pull/456#issuecomment-295810856 I'm really skeptical that you actually need this API and I'm fairly sure that including support for origin and/or domains at any level is just going to leak into other places in the API so we have to support it there as well. Unless you can show me a real use case that is really difficult for you right now I'm inclined to pass.

PR #468, opened 2017-04-24
Use jest-matchers for assertions instead of expect

skovhus:
Currently expect@1.x is being integrated into Jest (see more background in https://github.com/facebook/jest/issues/1679).

In [jest-codemods](https://github.com/skovhus/jest-codemods) I'm working on a codemod to automate the process of going from mjackson/expect to jest-matchers (if running in a browser) or plain Jest (if running in node). 🤖  

I've tested the codemod on this project and everything should still work. 

Let me know what you think.

## Update

**Updated 4th Septhember with Expect@21**


mjackson:
Hey @skovhus, thanks for taking a look at this. Do you have any updates here? I'd like to make this upgrade happen sometime soon. Thanks! :)

skovhus:
@mjackson would you like to still use a browser for executing these tests? If so then I can wrap this PR up using jest-codemods, so you can use the `jest-matchers`. At one point the `jest-matchers` package will be renamed to `expect`, but that should be a small change at that point. : )

skovhus:
@mjackson I've just reran the codemod on latest master. So I think this should be ready for a merge if you want.

mjackson:
Thanks, @skovhus. I need to keep using the browser for these tests because part of this library's job is to work around bugs and/or slightly differing behavior in various browser implementations, so the in-browser tests are crucial. I'd also like to make sure the ES5 build of the next version of `expect` works well in browsers. The current version of the package works well in browsers, and we need to preserve that if we want to provide a smooth upgrade path.

skovhus:
@mjackson I totally agree with the browser support for next version of expect/jest-matchers. 👍🏻 So while working on the codemod, I've also been contributing browser support to Jest:
- https://github.com/facebook/jest/issues/3360
- https://github.com/facebook/jest/pull/3421
- https://github.com/facebook/jest/pull/3476
- https://github.com/facebook/jest/pull/3532

We are just missing a few things it seems. : )

skovhus:
I'll update this once https://github.com/facebook/jest/pull/4367 has landed.

skovhus:
@mjackson master has been merged in now that expect@21 has been released. Should be ready for you to merge. : )

mjackson:
Thanks @skovhus! I'm taking a look at this now.

Issue #470, opened 2017-04-26
Why is a new location created when navigating to same location?


pshrmn:
This was brought up in https://github.com/ReactTraining/react-router/issues/5039, but is more appropriate here.

Basically, the problem is that if you are on page `/here` and click a link to `/here`, you end up with essentially duplicate (just different `key`s) location objects. Then, clicking the back button will bring you back to `/here`. This is obviously intentional because the [`PushSamePath`](https://github.com/ReactTraining/history/blob/master/modules/__tests__/TestSequences/PushSamePath.js) test sequence exists, but I'm not completely sure why.

However, that seems to differ from how browsers handle clicking on links to the same location. They seem to essentially do a `replaceState` when clicking a link that points to the same location (if you are on `/here`, click a link to `/other`, click the back button so that you are back on `/here` and click a link to `/here`, it does not wipe out the future location). Calling `window.location.href` with the same path also does not wipe out future locations, so I think that behavior is different for users who use the `forceRefresh` option.

I'm not sure what the ideal solution would be. I feel like it should be `history` that accommodates this.

For example, if it was left to other packages to adapt to this, `react-router-dom`'s `<Link>` would have to be able to detect when its `to` is the same as the current location and call `history.replace` instead of `history.push`. 

My initial instinct would be that `push` should detect when it is pushing the same location and call `replaceState` instead of `pushState`.

mjackson:
The `PushSamePath` test sequence exists only because it's not actually possible to push the same path using hash history (setting `window.location.href = window.location.href` will be ignored, i.e. does not fire `hashchange`) so that test was just ensuring that you actually *could* push the same path with browser history. Anyway, that only explains where the test came from. I'm not sure what the correct behavior is. Seems like the kind of thing that wouldn't be defined very well across browsers.

pshrmn:
From the [WHATWG browser spec](https://html.spec.whatwg.org/multipage/browsers.html#entry-update)

> If the navigation was initiated with a URL that equals the browsing context's active document's URL
> 1. Replace the current entry with a new entry representing the new resource and its Document object, related state, and the default scroll restoration mode of "auto".
> 2. Traverse the history to the new entry.

Where "equals" is defined [here](https://url.spec.whatwg.org/#concept-url-equals) (basically, `(a.pathname + a.search + a.hash) === (b.pathname + b.search + b.hash)`)

pshrmn:
[...]

PR #472, opened 2017-04-29
Add history.navigate API

pshrmn:
This is a proof of concept related to #470. This PR adds a `navigate` method to each history type.

This is inspired by the [WHATWG browser spec](https://html.spec.whatwg.org/multipage/browsers.html#update-the-session-history-with-the-new-page)

> If the navigation was initiated with a URL that equals the browsing context's active document's URL
> 1. Replace the current entry with a new entry representing the new resource and its Document object, related state, and the default scroll restoration mode of "auto".
> 2. Traverse the history to the new entry.

The basic chain of specs I used to find this started with the [`<a>` spec](https://html.spec.whatwg.org/multipage/semantics.html#the-a-element), which leads to the description of [how to follow hyperlinks](https://html.spec.whatwg.org/multipage/semantics.html#following-hyperlinks-2), which in turn leads to the description of [how browsers should navigate](https://html.spec.whatwg.org/multipage/browsers.html#navigate).

### `history.navigate(path, state)`

The `navigate` method has the same argument signature as `push` and `replace`.

When `history.navigate` is called, it creates a location object and compares it to the current location object. The only fields that are compared are: `pathname`, `search`, and `hash` (essentially comparing the serialized version of each location).

If the next location and the current location are the same, then a `REPLACE` action will occur. If they are different, then a `PUSH` action will happen (the actual implementation varies by history type).

```js
// when the current location = { pathname: '/fruit', search: '', hash: '#berries' }
history.navigate('/fruit#berries') // REPLACE
history.navigate('/fruit#citrus') // PUSH
history.navigate('/vegetables') // PUSH
```

#### Implementation

I added `createPushCallback` and `createReplaceCallback` functions inside each of the history types. These return the callback functions that were being passed to `transitionManager.confirmTransitionTo`. This approach was a lot easier than trying to play Frankenstein and putting together a monster of a callback inside of `navigate`.

`push` and `replace` always use their respective functions. `navigate` determines which one to use based on if the next location is the same as the current location.

### History Dependents (specifically, React Router)
With React Router, this would allow the `<Link>` to use `navigate` instead of `push`. Then, the rendered `<a>` will have the same behavior as a regular `<a>`. Potentially a `push` prop would need to be added to `<Link>` so that if a user _really_ wants to `push` the same location, they could, but I'm not sure why that would be.

```js
<Link to='/somewhere'>...</Link> // use history.navigate
<Link replace to='/still-here'>...</Link> // use history.replace
<Link push to='/go-there'>...</Link> // use history.push
```


modules/LocationUtils.js
```diff
@@ -59,3 +59,8 @@ export const locationsAreEqual = (a, b) =>
   a.hash === b.hash &&
   a.key === b.key &&
   valueEqual(a.state, b.state)
+
+export const locationPathsAreEqual = (a, b) =>
```
pshrmn: Names are hard :man_shrugging:

pshrmn:
Closing this since it is stale. I ended up writing my own history-esque package (inspired by this and the history from `vue-router`) for Curi that implements this functionality and it works well. It is nice because if you navigate to the same page that you are currently on, it doesn't wipe out "future history".

mjackson:
I recognize a lot of this code :D https://github.com/vuejs/vue-router/blob/dev/src/history/hash.js

mjackson:
@pshrmn Do you have a link to your work? I'd like to check it out. Thanks 💕

pshrmn:
https://github.com/pshrmn/hickory (an amalgamation of history and curi)

Most everything should be familiar to you. I changed some property names, but the but API is nearly the same. Biggest differences are that I use a different blocking system, I went HTML5 only, and (the main reason that I wrote a new package) I switched to `location.query` from `search` and added built-in parsing/stringifying (it just uses strings if the user does not provide their own). Perhaps the most jarring change is that there are semicolons everywhere :wink:.

It actually mostly works with React Router if you assign some property names that RR expects. https://codesandbox.io/s/gpg6lLo6 I wouldn't recommend that to anybody, just tested it out of curiosity.

Anyways, I'd be happy to give insight if you have any questions.

PR #473, opened 2017-05-04
createHref trailing slashes

robedmo:
Improves how `createHref` combines `basename` with `location`, particularly how it deals with trailing slashes.

The primary motivation for this change is an issue with the way `location.pathname == '/'` is currently handled; if the history has `basename == '/base/name'`, the result of `createHref({ pathname: '/' })` will be `/base/name/` with no option to get `/base/name` instead.

 Adds a new `createHref` function to the `PathUtils` module
 - Includes an options argument for determining whether to include the trailing slash in the case mentioned above (defaults to not include it). The options object can also include a property to indicate that the result should either have or not have a trailing slash in any case.
 - Currently, browser history is changed to use this function with the default options.
 - Intended to be used by [`StaticRouter`](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/modules/StaticRouter.js) in [`react-router`](https://github.com/ReactTraining/react-router) as well, so hrefs created by it will match. (example: server renders page with `StaticRouter` and then client renders it with a browser history)

This is a work in progress. TODO includes:
 - update hash history to use new `createHref` too.
 - add props to `createBrowserHistory()` and `createHashHistory()` options to determine which options will be used when `history.createHref(location)` calls `createHref(basename, location, trailingSlashOptions)`.
 - update `StaticRouter` component in `react-router` package to use the new `createHref` function.


robedmo:
Ran into the [100 push/replaceState limit on Safari 9](https://github.com/ReactTraining/history/issues/291) with all the new tests. That seems like a great decision on their part...

Working on reducing the number of state changes needed to run the tests.

robedmo:
While dealing with the Safari limit I noticed the absurd number of warnings that falsely claimed things like "/the/base/the/path" doesn't start with "/the/base".

I changed the condition for the warning
 - from `!(basename && hasBasename(path, basename))`
 - to `(!basename || hasBasename(path, basename))`

According to [the documentation](https://github.com/BerkeleyTrue/warning), the warning will be displayed when the condition is false; it represents something that should be true.

The test output is still filled with the warnings (because the Safari issue prevents each test from setting the window location before creating its history), but now they warn in the appropriate cases.

mjackson:
> The primary motivation for this change is an issue with the way location.pathname == '/' is currently handled; if the history has basename == '/base/name', the result of createHref({ pathname: '/' }) will be /base/name/ with no option to get /base/name instead.

Seems like this should require a fairly small code change. Why 18 commits?

mjackson:

PR #480, opened 2017-05-23
[v3] Expose `getCurrentIndex` method and add new `getKeysLength` method

### LINKED ISSUE #475: [v3] Expose `getCurrentIndex` and `allKeys.length`
Since `History.length` (#286) can't be easily implemented it will be great to expose `createHistory` child function `getCurrentIndex` and getter for `allKeys.length` on history object.
It will give an ability to developers to deal with some sort of history length and current position under own responsibility.

This issue relevant for v3 branch

gbiryukov:
Resolves #475 
By using this methods developers will be able to implement correct back button behaviour.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,8 @@
+## [HEAD]
+> May 23, 2017
+
+- Expose `getCurrentIndex` method and add new `getKeysLength` method that returns history keys count ([#475])
+
 ## [v3.3.0]
 > Feb 25, 2017
 
--- a/modules/createHistory.js
+++ b/modules/createHistory.js
@@ -34,6 +34,8 @@ const createHistory = (options = {}) => {
     return -1
   }
 
+  const getKeysLength = () => allKeys.length
+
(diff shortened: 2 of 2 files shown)


CHANGES.md
```diff
@@ -1,3 +1,8 @@
+## [HEAD]
+> May 23, 2017
+
+- Expose `getCurrentIndex` method and add new `getKeysLength` method that returns history keys count ([#475])
+
```
timdorr: Please don't add to the changelog in the PR. This will cause merge conflicts if anyone else does the same.

timdorr:
Unfortunately, this won't work at all if the user refreshes the page. `allKeys` is an internal construct for this `history` wrapper to track `location.key` uniqueness and doesn't correspond to `window.history.length`. This isn't something we can wrap, because there is the browser's history and the library's subset of that. You can't necessarily know the boundary line between the two.

gbiryukov:
clear. thanks for the feedback

PR #485, opened 2017-06-06
Fix logic bug when displaying history basename warning messages

jeffreypriebe:
Previously, if "there is a basename" but the path "has the basename" the warning would show.

Now correctly doesn't show the warning.

### DIFF
--- a/modules/createBrowserHistory.js
+++ b/modules/createBrowserHistory.js
@@ -60,7 +60,7 @@ const createBrowserHistory = (props = {}) => {
     let path = pathname + search + hash
 
     warning(
-      !(basename && hasBasename(path, basename)),
+      (!basename || hasBasename(path, basename)),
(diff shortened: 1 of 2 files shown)


pshrmn:
Yeah, this should be merged. `warning` strikes again. It appears I broke this in #459 by not negating the `hasBasename` result :man_facepalming: . At least there hasn't been a release since that was merged.

jeffreypriebe:
@pshrmn The '`false` = show` of 'warning' makes it so I had to read it _very_ _carefully_ to get it right. I understand why they choose that way, but it is counter intuitive.

timdorr:
Merging as a bugfix, as per https://github.com/ReactTraining/history/pull/465#issuecomment-307900191

mjackson:
Thanks for this, @jeffreypriebe 💃

PR #486, opened 2017-06-06
Add path-to-regexp style basename

jeffreypriebe:
Allow path parameters in the `basename` (e.g. `basename` of '/never/:param' will remove '/never/gonna' from path '/never/gonna/give').

Using same package ('path-to-regexp') as react-router for its matchPath (so the behavior should be the same as the `path` prop of `<Route />`).

Tests, README.md, and CHANGES.md updated.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,8 @@
+## 
+> Jun 6, 2017
+
+- Add path-to-regexp to the `basename` parsing, allowing use of parameters in `basename`
+
 ## [v4.6.1]
 > Mar 15, 2017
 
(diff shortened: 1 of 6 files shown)


jeffreypriebe:
While this works for matching in a simple form, it has various problems with other portions of the history API. Closing for now.

austenpayan:
did anything become of this? I would love this functionality

jeffreypriebe:
@pshrmn Agreed - exactly as you outline is what I ended up doing.

@austenpayan You can see what I had previously cobbled together over here: https://github.com/jeffreypriebe/history/tree/feature/strip-prefix-path-params
I like the basepath working in the same way as routes do (with path-to-regexp) but as @pshrmn notes, it's easy to do it yourself and then there is no confusion about the params.

Issue #489, opened 2017-06-09
confirmTransitionTo prompt require type string, maybe it shouldn't?


fernandojbf:
Hi all,

This may not make sense but maybe we could pass `prompt` as an Object? With this we can configure different outcomes of the `getUserConfirmation` function. Any thoughts about this or does it not make any sense?

Cheers

TimoSta:
We are facing the same issue right now.

We'd like to have a multi-option confirmation popup (save/discard/cancel) and need a way to pass additional details about the actions, and an action handler into `getUserConfirmation`. This is currently not possible.

Any thoughts on this? More than happy to submit a PR if there is interest.

mjackson:
We are removing `history.block` (it's buggy, see #690), so this won't be an issue for us anymore.

Issue #483, opened 2017-05-29
support for promise-based block?


moonforeshot:
I'm thinking about customizing block operations on a per-individual basis.
For example, one part of code requires Auth API promise, another would require Item API promise.

What do you think about supporting individual-level customizable block API?

Example code:

```javascript
history.block(async (location, action) => {
     ...
     try {
         const loggedIn = await getLogin( ... );
         return true;
     } catch { return false; }
})
```

Moreover, it might be better to have a generalized 'customized-catch-and-redirect':

```javascript
history.beforeChange(async (location, action, callback) => {
   
    const userLevel = await getUserLevel( ...);
    if (userLevel == "admin") { 
         return location
    }
    else if (userLevel == "user") { 
         return modifiedLocationForNotAdmins; 
     }
     else {
          return false; //block.
     }
});
```

mjackson:
Sure, I'd be open to including support for promises in the blocking API. Just needs a PR.

PR #493, opened 2017-06-14
Implements new API for blocking transitions with generic pre-location-change hook

lourd:
A new `before` method allows functions to register as pre-transition hooks that get called before a change in history's state. These methods can be async by returning a promise.

This new pre-hook functionality simplifies the API for conditional or asynchronous blocking of a location transition. It removes the responsibility of determining how to ask a user for their confirmation from the library and leaves it up to the application.

History instances no longer have a `block` method and the factory functions no longer receive a `getUserConfirmation` option.

The API allows for multiple pre-hooks to be registered, so I think it can also resolve the dispute over including query parsing & stringifying, since that feature can simply be added as a pre-transition hook.

~~I also organized and simplified the tests, and added expanded coverage on expectations about the window's state vs the history's state when blocking transitions asynchronously. I changed the async primitive of TestSequence files and `execSteps` to be a promise instead of a callback. This allows for async functions to be used, which is much simpler to follow and catch errors for. Additionally, I reworked the TestSequence files to include their `describe` and `it` blocks
to cut down on the repeated test descriptions.~~

~~One significant tradeoff to the hook implementation is that it now requires the library to have `babel-runtime` as a peer dependency in order to run the transpiled version of the `async` function and be sure of the global `Promise` availability. The implementation is certainly the most succinct way to iterate through an array asynchronously with bailout functionality, but I acknowledge babel-runtime is not the lightest dependency. I'm happy to discuss other options.~~

~~I also made the unlisten functions idempotent. I didn't realize #440 was already
open for this same behavior. This implementation uses one less variable,
but mutates an argument, which can be a questionable pattern but I think
is fine here~~

I think this covers #483, #484, & #489

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,24 +72,18 @@ createBrowserHistory({
   basename: '',             // The base URL of the app (see below)
   forceRefresh: false,      // Set true to force full page refreshes
   keyLength: 6,             // The length of location.key
-  // A function to use to confirm navigation with the user (see below)
-  getUserConfirmation: (message, callback) => callback(window.confirm(message))
 })
 
 createMemoryHistory({
   initialEntries: [ '/' ],  // The initial URLs in the history stack
   initialIndex: 0,          // The starting index in the history stack
   keyLength: 6,             // The length of location.key
-  // A function to use to confirm navigation with the user. Required
-  // if you return string prompts from transition hooks (see below)
-  getUserConfirmation: null
 })
 
 createHashHistory({
   basename: '',             // The base URL of the app (see below)
   hashType: 'slash',        // The hash type to use (see below)
   // A function to use to confirm navigation with the user (see below)
-  getUserConfirmation: (message, callback) => callback(window.confirm(message))
 })
 ```
 
@@ -103,7 +97,9 @@ Each `history` object has the following properties:
 
 Additionally, `createMemoryHistory` provides `history.index` and `history.entries` properties that let you inspect the history stack.
 
-### Listening
+### Transition Hooks
+
+#### Post-change hook
 
 You can listen for changes to the current location using `history.listen`:
 
@@ -113,6 +109,45 @@ history.listen((location, action) => {
   console.log(`The last navigation action was ${action}`)
 })
 ```
+
+Post-change listener functions will be not be invoked when a transition is blocked by a pre-change hook, unless the user used the built-in browser back or forward features to initiate the transition. In this case, the listener functions will be called with the same `location` and `action` as the previous invocation.
+
+#### Pre-change hook
+You can perform side effects before history state changes by registering functions with `history.before`
+
+```js
+const unsub = history.before((location, action) => {
+  location.query = queryParseFn(location.search)
+})
+```
+
+Before hooks can also be `async` functions, i.e. return a Promise, to block the state change asynchronously.
+
+```js
+const unsub = history.before(async (location, action) => {
+  await getDataAndCodeForLocation(location)
+})
+```
+
+Returning a truthy value from any before hook functions will cancel the transition.
+
+```js
+const unsub = history.before((location, action) => {
+  return await askUserIfTheyreSureTheyWantToLeave()
+})
+```
+
+Or in a simpler, less custom fashion
+
+```js
+const unsub = history.before(async (location, action) => {
+  return window.confirm('Are you sure you want to leave this page?')
+})
+```
+
+The hooks are called in the order they're registered. Not every hook will be called if one of them blocks the transition.
+
+### `Location`
 
 The `location` object implements a subset of [the `window.location` interface](https://developer.mozilla.org/en-US/docs/Web/API/Location), including:
 
@@ -166,46 +201,6 @@ history.goBack()
 ```
 
 **Note:** Location state is only supported in `createBrowserHistory` and `createMemoryHistory`.
-
-### Blocking Transitions
-
-`history` lets you register a prompt message that will be shown to the user before location listeners are notified. This allows you to make sure the user wants to leave the current page before they navigate away.
-
-```js
-// Register a simple prompt message that will be shown the
-// user before they navigate away from the current page.
-const unblock = history.block('Are you sure you want to leave this page?')
-
-// Or use a function that returns the message when it's needed.
-history.block((location, action) => {
-  // The location and action arguments indicate the location
-  // we're transitioning to and how we're getting there.
-
-  // A common use case is to prevent the user from leaving the
-  // page if there's a form they haven't submitted yet.
-  if (input.value !== '')
-    return 'Are you sure you want to leave this page?'
-})
-
-// To stop blocking transitions, call the function returned from block().
-unblock()
-```
-
-**Note:** You'll need to provide a `getUserConfirmation` function to use this feature with `createMemoryHistory` (see below).
-
-### Customizing the Confirm Dialog
-
-By default, [`window.confirm`](https://developer.mozilla.org/en-US/docs/Web/API/Window/confirm) is used to show prompt messages to the user. If you need to override this behavior (or if you're using `createMemoryHistory`, which doesn't assume a DOM environment), provide a `getUserConfirmation` function when you create your history object.
-
-```js
-const history = createHistory({
-  getUserConfirmation(message, callback) {
-    // Show some custom dialog to the user and call
-    // callback(true) to continue the transiton, or
-    // callback(false) to abort it.
-  }
-})
-```
 
 ### Using a Base URL
 
--- a/modules/__tests__/BrowserHistory-test.js
+++ b/modules/__tests__/BrowserHistory-test.js
@@ -125,20 +125,20 @@ describeHistory('a browser history', () => {
       })
     })
 
-    describe('block', () => {
+    describe.skip('block', () => {
       it('blocks all transitions', (done) => {
         TestSequences.BlockEverything(history, done)
       })
     })
 
-    describe('block a POP without listening', () => {
+    describe.skip('block a POP without listening', () => {
       it('receives the next location and action as arguments', (done) => {
         TestSequences.BlockPopWithoutListening(history, done)
       })
     })
   })
 
-  describe('that denies all transitions', () => {
+  describe.skip('that denies all transitions', () => {
     const getUserConfirmation = (_, callback) => callback(false)
 
     let history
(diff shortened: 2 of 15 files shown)


lourd:
The tests can be a little flaky 😒 struggled a bunch with the ones trying to make calls to the actual `window.history` and wait for resulting events to propagate before asserting something. They're passing consistently on my machine, but looks like not so much on the browser stack build.

modules/__tests__/BrowserHistory-test.js
```diff
@@ -11,187 +12,33 @@ describeHistory('a browser history', () => {
...
-
-    it('is called when the back button is clicked', (done) => {
-      TestSequences.BackButtonTransitionHook(history, done)
-    })
-
-    it('is called on the hashchange event', (done) => {
-      TestSequences.HashChangeTransitionHook(history, done)
-    })
+  describe('by default', () => {
+    generateTests(cases, createHistory)
```
mjackson: There's nothing "cleaner" about using a loop to generate the test cases. Besides the fact that the code is now less explicit, it's also more difficult to use `describe.only` and `it.only` in development to run only a single test or suite.

lourd:
Yeah, I got a little carried away with the changes. Sorry about that.

I decided to start fresh, instead of trying to backtrack. This branch now has just the minimal set of and an update to the README for explanation. We can discuss the API first then I can re-create the relevant tests.

pshrmn:
What is the rationale for having multiple `before` calls? Is it just so that you can use `before` to manipulate the `location`? I want to say that that is the case because I'm having trouble coming up with why you would have multiple navigation blocks.

mjackson:
Can you say a little about why you prefer `before` instead of `block` + `getUserConfirmation` @lourd? Seems like more work.

bwhitty:
@lourd can you comment on how this would be different or better than https://github.com/ReactTraining/history/pull/484? Your solution introduces breaking changes which don't seem to be necessary. I think #484 achieves everything you want in a simple, backward compatible way.

@mjackson / @pshrmn could you possibly weigh in on this thought?

lourd:
@mjackson, surely. Basically `getUserConfirmation` is only configurable at the time of history creation, and `block` is inflexible. It only allows one "block", and it is difficult to use in coordination with `getUserConfirmation`, which is required to fulfill the use case some folks are looking for, such as:

1. computing something asynchronously before deciding whether to block or not
2. showing different types of transition confirmation UIs depending on the location in the application
3. blocking the new location state from propagating into the UI while fetching the new location's data.

I've had to accomplish the first two cases in projects, which is where I came up with this API. The third case is something I've observed folks in the Redux community discussing — basically managing location transitions as side effects. See [redux-tower](https://github.com/kuy/redux-tower) and [redux-first-router](https://github.com/faceyspacey/redux-first-router) for other, much more complex libraries aimed at tackling that API challenge.

The explanations that @bwhitty [gave](https://github.com/ReactTraining/history/pull/484#issuecomment-309349640) and @karagoth [gave](https://github.com/ReactTraining/history/pull/484#issuecomment-309411662) over in #484 also elucidate some of the alluded pain points pretty well.

@pshrmn, I agree that the majority of folks won't have a need for more than one pre-transition hook, but some will, and the `before` API supports them, where the `getUserConfirmation` API does not. I think having a slightly lower-level API of a pre-state-change hook, basically a middleware for location transitions, is easier to explain and use then the current API. I think the fact that the library has to concern itself with the UI at all (`window.confirm`) is code smell.

As far as breaking changes, or which async primitive to use (callbacks instead of Promises?), or method names go (leave the name as `block` instead of `before`?) — I'm happy to discuss and compromise. It is possible to make a version of pre-transition hooks that isn't a breaking change with the current functionality of `getUserConfirmation` + `block`, or at least "less" of a breaking change.

lourd:
[Here's a gist](https://gist.github.com/lourd/8fc409781667bb25c0e6cd95520ed349) I quickly whipped together for what this API could look like from an application standpoint.

After reading through https://github.com/ReactTraining/react-router/issues/4635#issuecomment-290415863, it's clear folks are stumbling on this use case with `react-router` + `history`.

bwhitty:
Ping @mjackson (sorry to do that). I think we've put forward some legitimate use-cases this or #484 solves. Could you provide some feedback on whether this is something you're willing to support?

bwhitty:
Hello @pshrmn @mjackson, any indication of the receptiveness to this approach would be appreciated. I'd love to help in any way to move this feature set forward.

lourd:
Hey y'all,

Hope you've had a good summer. Been a while since we had any feedback on this PR. Where do we stand?

I was reading through the Next.js source and [found another references to point to for the need for async history-changes](https://github.com/zeit/next.js/blob/master/lib/router/router.js#L154-L182). Next's async abstractions are a little higher level, with the mapping of URLs to component loading and data fetching built into the framework, but the proposed changes in this PR would enable similar functionality to be built using `history` and `react-router`.

lourd:
Friendly monthly bump.

README.md
```diff
@@ -114,6 +110,45 @@ history.listen((location, action) => {
...
+const unsub = history.before((location, action) => {
+  location.query = queryParseFn(location.search)
+})
+```
+
+Before hooks can also be `async` functions, i.e. return a Promise, to block the state change asynchronously.
+
+```js
+const unsub = history.before(async (location, action) => {
+  await getDataAndCodeForLocation(location)
```
mjackson: What is the user looking at while we wait for `getDataAndCodeForLocation` to complete? The URL has changed, but they're still looking at the old page because history hasn't notified anyone. :/

README.md
```diff
@@ -114,6 +110,45 @@ history.listen((location, action) => {
...
 ```
 
+Post-change listener functions will be not be invoked when a transition is blocked by a pre-change hook, unless the user used the built-in browser back or forward features to initiate the transition. In this case, the listener functions will be called with the same `location` and `action` as the previous invocation.
+
+#### Pre-change hook
+You can perform side effects before history state changes by registering functions with `history.before`
+
+```js
+const unsub = history.before((location, action) => {
+  location.query = queryParseFn(location.search)
```
mjackson: This use case is legit for query parsing, but I can't think of many other use cases. This use case could be solved in a lot of other ways though, like just wrapping all your `listen` handlers before you use them.

```js
function parseQuery(callback) {
  return function (location) {
    if (!location.query)
      location.query = parseTheQueryString(location.search)

    callback(location)
  }
}

history.listen(parseQuery(callback))
```

README.md
```diff
@@ -114,6 +110,45 @@ history.listen((location, action) => {
...
+const unsub = history.before(async (location, action) => {
+  await getDataAndCodeForLocation(location)
+})
+```
+
+Returning a truthy value from any before hook functions will cancel the transition.
+
+```js
+const unsub = history.before((location, action) => {
+  return await askUserIfTheyreSureTheyWantToLeave()
```
mjackson: This use case is solved by `history.block`, or at least it's the use case that prompted that API.

README.md
lourd: Exactly. The user still has the ability to see the current page, with all of its loaded data and information, while the new one loads in the background. The view only switches once the data is available and the page can be shown. The user can be shown some kind of progress indicator while that's happening.

This design can be desirable on slower networks where there's a significant amount of latency. Instead of having my users stare at a blank page with spinners or placeholder UI, I can allow them to keep looking at the data they already have. This is the behavior of Chrome and most browsers, I believe. Clicking on a link for a new document will leave the previous document there until any parts of the new document have been received.

lourd:
Hey @mjackson, you're welcome. Thank you for your reply.

I understand that creating an async pre-change hook opens up the possibility for poor app design if people don't communicate the happenings to their user. I don't agree that that's a good reason to restrict the the functionality of the library, though. That can be resolved with guidance and messaging in documentation and examples.

I believe we've illustrated valid shortcomings with the current API. The scroll opt-out is still an unresolved use case, and the current way to use `getUserConfirmation` folks have found confusing — having the UI state management coupled to the `getUserConfirmation` parameter at `history` creation, rather than being able to configure it later on at the component doing the blocking.

README.md
mjackson: > The user can be shown some kind of progress indicator while that's happening.

No, you won't be able to do this because you won't know the location has changed. The user could be typing into the address bar or clicking the back button (i.e. *not* clicking on one of your links, and therefore out of your control), in which case you won't be notified the location has changed at all. Imagine someone hitting the back button and then waiting a few seconds before anything happens. They would think the site is broken.

It's not history's job to lie to you and pretend the location hasn't changed when it has. This doesn't mean that you're not free to show the user some old UI while you load some data. You absolutely can. Just look at the way React's transition group works. It gets new props (children), and gradually fades the old ones out. But you can't fade the new ones in unless you know what they are first! You can think about this library in the same way. If you're making a transition from one location to another, you will probably need to know what the destination is in order to do anything meaningful.

I hope that made sense. I'm sorry, but I don't think this is the right approach.

lourd: You would know the location has started to change from the function that was registered with `before`. That function can make the calls it needs to change state to show a loading indicator or do whatever else.

> If you're making a transition from one location to another, you will probably need to know what the destination is in order to do anything meaningful.

You do know the destination. It's the argument given to the functions registered with `history.before`.

lourd:
It's disappointing to see this closed without addressing the unsolved scrolling opt-out use case or the shortcomings of the current `getUserConfirmation` + `block` API. The scrolling case was why I started work on this in the first place 😪

Your arguments against the nature of location changes being synchronous or asynchronous are especially confusing since the current API already "lies" about the location when doing async work in `getUserConfirmation`.

mjackson:
I agree that `getUserConfirmation` is probably not the best API. I explained why I originally used that API in https://github.com/ReactTraining/history/pull/484#issuecomment-329684203, which I think is probably a better way to go to address your use cases. Again, I'm sorry I had to close but I'm not convinced this is the right approach.

lourd:
Your proposed API in [#484 (comment)](https://github.com/ReactTraining/history/pull/484#issuecomment-329684203)

> `history.block((location, action, callback) => ...)`

is essentially the same as this one, just with a different method name and async mechanism.

mjackson:
The use case for `history.block` is to prevent users from navigating away from the current page, not to do some async work in preparation for showing the next page. The only reason it's async is because people may want to show some custom UI to get the user confirmation instead of using something sync like `window.confirm`. But yes, that API could potentially be used to do what you're proposing with `history.before`, but I'd discourage anyone from using it like that.

PR #495, opened 2017-06-18
Add main/module entries

pshrmn:
I didn't realize that these fields were missing until I came across an issue with using history on CodeSandbox https://github.com/CompuIves/codesandbox-client/issues/22.


mjackson:
Ha, that's nuts! I guess now we know nobody is using the ES6 modules build ;) Thanks, @pshrmn!

PR #498, opened 2017-06-23
Update PathUtils.js

theCuriousOne:
Conflict with other plugins (ex. source-map), which exports a function named "search" and this part of the script crashes, because it is not known search.charAt() (search is a function).

--All test have been successfully completed (I don't believe a new test is needed)
--ESlint shows no breakage in code

### DIFF
--- a/modules/PathUtils.js
+++ b/modules/PathUtils.js
@@ -42,10 +42,10 @@ export const createPath = (location) => {
 
   let path = pathname || '/'
 
-  if (search && search !== '?')
+  if (search && (typeof search === 'string') && search !== '?')
     path += (search.charAt(0) === '?' ? search : `?${search}`)
 
-  if (hash && hash !== '#')
+  if (hash && (typeof hash === 'string') && hash !== '#')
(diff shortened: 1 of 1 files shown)


mjackson:
The API here is `createPath({ pathname: '', search: '', hash: '' })`. `createPath({ search: () => {} })` is not supported.

theCuriousOne:
I know that (now) @mjackson . I had the mistake when migrating from an old version of history (v3) to newer one (v4+). I am not saying it is a bug in the code, but since in the code there is already the same logic implemented, why not here also? On the other hand it would had help me narrow the bug faster, so it might be useful for others as well.

mjackson:
I don't know what "plugins" you're referring to, or what you mean by "the same logic". Same as what?

theCuriousOne:
@mjackson "the same logic" means that you also you in the code the checking of the type in the if condition (for example` if (typeof path === 'string')` in Location) or you can just search the repository for "typeof" to see what I have meant. 

Disregard plugins, it is confusing to explain, but in short "I had some problems when learning this library, and the errors in code misdirected me".

Sorry to bother you.

PR #506, opened 2017-07-26
Fix sessionStorage leak

aij:
Use history state to store the state directly, rather
than indirecting into sessionStorage which was never
freed until the session was closed.


mjackson:

aij:
Yes, having our apps not randomly crash after a while is still important... We thought it might be important to others as well.

If it helps, we've been using this in production for quite a while now.

Issue #516, opened 2017-08-28
Was automatic decoding a mistake?


pshrmn:
Since there have been issues, I thought that it might be appropriate to discuss the pros and cons of automatic `pathname` decoding in `history`.

#### Pros

* Consistent `location` objects. If `history` always decodes the pathname, then location object's will always have a decoded `pathname`. If `history` does not do this, then `location`s created from `window.location` will have an encoded pathname, while `location`s created from the user will be encoded/unencoded based on what the user provides.
* Unencoded `pathname`s make route matching easier (for React Router).

#### Cons

* Makes it difficult if the user wants the `pathname` to be encoded. This can be because of special handling (like the encoded forward slash https://github.com/ReactTraining/history/issues/505) or because of illegal encoding (https://github.com/ReactTraining/history/issues/481) that some websites (like Google) still support.

**Edit:** Deleted some stuff here, nothing particularly important.

Perhaps I'm just overthinking this, but whenever I see issues with the current implementation, it makes me think I jumped the gun when I implemented the automatic decoding. It definitely seems like a good idea, but text encoding is difficult/annoying to deal with and I think that it might be better to have consistent behavior than convenience with some edge cases.

pshrmn:
I just took a look at what Django does, and they also decode pathnames (to the extent that a request for `/test%2fing` will be matched by `r"^test/ing$"`, which I'm not entirely convinced is the desirable). They also reference [RFC 3987](https://www.ietf.org/rfc/rfc3987.txt), which I should probably take a look at (not sure how relevant it is for `history`). 

I think that the important thing is that the `pathname` pushed to the browser is what the user provides us. That should be doable while still automatically decoding the `pathname`. I'll look into a possible PR for this behavior sometime soon.

PR #517, opened 2017-08-29
Add rawPathname to location

### LINKED ISSUE #505: Path is decoded in createLocation
This results in an invalid URL being pushed to the HTML5 history element when there are encoded percent signs in the pathname.

E.g.:
```js
history.push({
  pathname: `randomlocation/${ encodeURIComponent("AB%C") }/furthermore`
});
```
results in `"randomlocation/AB%C/futhermore"` being pushed to the HTML5 history.

pshrmn:
This is a followup to #516 and fixes #505. I don't know if this is the best solution, but I thought that I'd put it out there. I implemented the same thing in [hickory](https://github.com/pshrmn/hickory) and it does its job well.

This property is set prior to decoding the `pathname`. For browser and memory histories, the `rawPathname` will be the encoded version of the provided `pathname` (generated using an anchor element). For memory histories, the `rawPathname` is the same as the `pathname`.

When creating a URL (using `createPath`), the `rawPathname` property will be preferred over the `pathname` property.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,7 @@
+## HEAD
+
+- Add `rawPathname` property to location objects and prefer that over `pathname` in `createPath`. For browser and hash histories, `rawPathname` is an encoded string, while for memory histories, it is the same as the `pathname`.
+
 ## [v4.6.3]
 > Jun 20, 2017
 
--- a/modules/DOMUtils.js
+++ b/modules/DOMUtils.js
@@ -56,3 +56,9 @@ export const supportsGoWithoutReloadUsingHash = () =>
 export const isExtraneousPopstateEvent = event =>
   event.state === undefined &&
   navigator.userAgent.indexOf('CriOS') === -1
+
+export const browserEncode = pathname => {
+  const a = document.createElement('a')
+  a.href = pathname
+  return a.pathname
+}
--- a/modules/LocationUtils.js
+++ b/modules/LocationUtils.js
@@ -2,7 +2,9 @@ import resolvePathname from 'resolve-pathname'
 import valueEqual from 'value-equal'
 import { parsePath } from './PathUtils'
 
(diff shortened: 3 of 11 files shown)


mjackson:
I'd really, really like to avoid adding `rawPathname` if we can avoid it. I've been holding off on tackling the encoding issues for too long now, hoping they would resolve themselves. But maybe it's time to dig in so I can understand why you think we need this.

pshrmn:
It has been a minute since I've thought about this, so I'm going to have to refresh my memory. I realize that the explanation is lacking (I'm pretty sure that Tweet was about this); I'll close this for now and only re-open if I think I can put forth a solid reasoning why this is necessary.

There is definitely part of me that regrets making that original decoding PR. I think that having `pathname`s be decoded is beneficial (otherwise there would be a mix of encoded and decoded pathnames depending on how the location is created and React Router would probably have to handle these problems instead), but it opens up all these little issues that remind you why text is hard.

parkercluff:
We ran into an issue with % being decoded, but other characters not being decoded. Similar to issue 505.

We were calling encodeURIComponent on pieces of our path for use in our Link components. If our entity name was bounce% it would encode to bounce%25, but the pathname being decoded in LocationUtils.js would decode the bounce%25 to bounce% and so when we ran that through decodeURIComponent on our end it breaks for malformed URI. 
If the name of the entity was bounce$ then it would encode to bounce%24 and would decode in LocationUtils.js to bounce%24, so when we locally use decodeURIComponent it would work perfectly and result in bounce$.

It seems that every character is decoding properly except for %.

If rawPathname was provided then the decoding could be handled on our end.

Also, a work around we found is if you add
`location.pathname = location.pathname.replace(new RegExp('%25', 'g'), '%2525');`
just before 
`location.pathname = decodeURI(location.pathname);`
in LocationUtils.js the issue seems fixed.

Thoughts?

Issue #511, opened 2017-08-09
PhantomJS pageloads generate an extraneous POP event


shepmaster:
I'm noticing extra `POP` events generated when running tests inside of PhantomJS. Some searching has found various reports of the same thing happening on older WebKit versions, which makes sense, as that's what Phantom is based on. 

The [current implementation of `isExtraneousPopstateEvent`](https://github.com/ReactTraining/history/blob/8d98aec1366b86e3832dbfd7dcc480d1f9a5275c/modules/DOMUtils.js#L56-L58) does not appear to have anything *specific* to PhantomJS:

```js
export const isExtraneousPopstateEvent = event =>
  event.state === undefined &&
  navigator.userAgent.indexOf('CriOS') === -1
```

I added some prints to this code and can see that `event.state` is `null`, not `undefined`.

In decidim/decidim#1458, they chose to use the `isTrusted` property, so that might be a direction? I see that the event only has these properties (`state,returnValue,timeStamp,eventPhase,target,defaultPrevented,srcElement,type,clipboardData,cancelable,currentTarget,bubbles,cancelBubble`), so maybe even just checking for the presence of that property would be enough.

For reference, my Phantom's user agent is `Mozilla/5.0 (Macintosh; Intel Mac OS X) AppleWebKit/538.1 (KHTML, like Gecko) PhantomJS/2.1.1 Safari/538.1`.

shepmaster:
As a local workaround, I've wrapped the history object and am passing it to redux-little-router:

```js
import { createBrowserHistory } from 'history';

// On page load, PhantomJS triggers two POP actions, while the newer
// spec says to only generate one. This causes double loads in our
// tests. This is inefficient and causes tests to fail when the DOM is
// loaded once, we start to interact with it, then the second load
// finishes and reverts our changes.
//
// Until this is fixed upstream, we ignore the first POP action that
// occurs if it looks like we are inside PhantomJS.
export default function createBrowserHistorySafeForTesting(args) {
  const history = createBrowserHistory(args);
  const originalListen = history.listen;

  let popCallsToBurn = howManyPopCallsToBurn();

  history.listen = realCallback => {
    const callback = (location, action) => {
      if (action === 'POP') {
        if (popCallsToBurn === 0) {
          realCallback(location, action);
        } else {
          popCallsToBurn--;
        }
      } else {
        realCallback(location, action);
      }
    };

    originalListen(callback);
  };

  return history;
}

function howManyPopCallsToBurn() {
[...]

PR #520, opened 2017-09-07
Normalize browsers that fire a popstate event on load

hydnhntr:
> Browsers tend to handle the popstate event differently on page load. Chrome (prior to v34) and Safari (prior to 10.0) always emit a popstate event on page load, but Firefox doesn't.

— [MDN: WindowEventHandlers.onpopstate](https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onpopstate)

Modern browsers that fired `popstate` on load no longer do so. Yahoo’s [fluxible-router](https://github.com/yahoo/fluxible/tree/master/packages/fluxible-router) features a wrapper around the History Web API and [normalized this behaviour across browsers](https://github.com/yahoo/fluxible/issues/349) by suppressing the event.

I’m not sure about the placement of my new additions, feel free to advise how I could integrate them better.

As Safari/WebKit exhibited this quirk #511 might be solved with these changes.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,10 @@
+## [v4.7.3]
+> Sep 7, 2017
+
+- Normalize browser behavior by ignoring popstate on load present in Chrome (prior to v34) and Safari (prior to 10.0).
+
+[v4.7.3]: https://github.com/ReactTraining/history/compare/v4.6.3...v4.7.3
+
 ## [v4.6.3]
 > Jun 20, 2017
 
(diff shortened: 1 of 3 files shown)


DevSide:
:+1:

mjackson:
Can you please say more about why you need this? I don't want to call `replaceState` on the initial page load if we can avoid it. We should be able to initialize a history object using the current state of the world and not have to change it immediately to conform to our expectations.

Ben07:
when would you merge this submit？I face the same problem

danywheeler:
I do have this problem in IOS below 9.3, the popstate fires on page load.

mjackson:
Closing due to lack of response from the OP.

PR #522, opened 2017-09-16
add basename option to memoryHistory

faceyspacey:
tests included etc.


faceyspacey:
credit goes to @e6exp for the `basename` implementation.

etburke:
This is great. Much needed.

mjackson:
Can you please elaborate on why you need a basename with memory history? The reason we need it in browsers is because you can't always control the URL in browsers. For example, you may need to "mount" your HTML app at a sub-path on your domain. But there is no such restriction when using memory history. You're in control of the entire URL the whole time.

re6exp:
When we use SSR from under gateway application (e.g. Spring Cloud Netflix Zuul/Gateway), we need to forward the connection to corresponding node.js server based on the input application path. In this case (with our patch we applied) SSR works perfect under cloud conditions. So, at the server-side, there can be some situations, when using basename with memory history is applicable.

mjackson:
If you're using history on the server, you're probably using React Router 3 or lower. In React Router version 4 we don't use history on the server at all.

If that's the case, this PR should probably target the v3 branch.

Michael-M-Judd:
@mjackson So are you removing `basename` support just to keep the library lean? IMO a basename is just a nice convenience factor. We have a unique micro-app setup where an entire repo is injected at some entrypoint within a larger app wrapper, and it always works off of _some_ basename, so it's nice to have the ability to not worry about the total URL all the time.

alfaex:
I need basename for the kiosk app that I maintain.
Since is a public app running in chrome kiosk mode, memory router allow me reset the navigation history when the client reach the end of the app workflow.

Kiosk mode on chrome don't have address bar, this is great because the user will never be able to reach the configurations url, but since memory router don't read from address bar the staff can't reach the config screen opening a normal window of chrome and navigating to myapp/config.

But what all this have to do with basename...

A solution that I found is use `initialEntries` and set it up with `window.location.pathname` so on a normal chrome window I can type myapp/config and that page will load, and on the kiosk mode there is no address bar so the root object it will be loaded.

But some versions run on subpath domain `myapp/kiosk/config` since i can't use basename the app get lost and render blank page. 

The location will print `myapp/kiosk/config` and on the app is listening for `myapp/config` so I need to generate a specific version for the clients that will work with subpaths domains.

Thanks.

Issue #502, opened 2017-07-03
location.key is undefined after createBrowserHistory


aristidesfl:
After createBrowserHistory(), location.key is undefined. This prevents location.key from being used to identify history items where createBrowserHistory() is used.

pshrmn:
Browser history could probably just call `replaceState` before creating the initial location when there is no key detected. @aristidesfl if you're interested in making a PR for this, it should be pretty simple.
```js
if (canUseHistory) {
  const currentState = getHistoryState()
  if (!currentState.key) {
    const newState = Object.assign({}, currentState, { key: createKey() })
    globalHistory.replaceState(newState, null, window.location.href)
  }
}
const initialLocation = getDOMLocation(getHistoryState())
```
The above code is probably more cautious than it needs to be. I can't think of a reason why `window.history.state` would be defined while `window.history.state.key` would not be (for an application that uses the `history` package). The following is probably enough:
```js
if (canUseHistory && !globalHistory.state) {
  globalHistory.replaceState({ key: createkey() }, null, window.location.href)
}
```

As far as testing it goes, you would just push a location with null state (`window.history.pushState(null, null, '/somewhere')`), create a browser history, and verify that `browserHistory.location.key` is defined.

Of course, whether this gets merged is up to @mjackson, but I think that it would be useful.

mjackson:
I used to call replace immediately, but it caused issues in some versions of IE. Can you please explain why you need a `key` on the initial location? Isn't it enough to know that on the initial load the `key` isn't defined?

mjackson:
Closing this one since conversation is stalled.

aristidesfl:
@mjackson I'm using the key for custom scroll position management. I could use `undefined` as a key, but then it overlap with other initial history sessions. Even if I make it work it will be a workaround. If it is a history position, shouldn't it have a key like the other ones?

aristidesfl:
The current workaround is doing a replaceState(), but it is also less than ideal.

PR #523, opened 2017-09-18
Ensure BrowserHistory provides initial key

ruggeri:
I've implemented the suggestion of @pshrmn with regard to addressing #502. I also updated the appropriate test.

Thanks humans of history!


mjackson:
After thinking about this for a while, I'm pretty sure we don't want to do a `replaceState` on the initial page render w/out some sort of explicit user opt-in. Thanks for the PR though, @ruggeri

PR #527, opened 2017-09-25
Fix broken link in v2 Glossary.md

emp823:
Link was to master/docs but the master branch does not have a docs folder.


mjackson:
We don't maintain v2 anymore, sorry. But thank you!

PR #528, opened 2017-09-25
Reference v3 Location docs in Glossary.md

emp823:
Link was to master but master does not have a docs folder


mjackson:
Thanks, @emp823 😄

PR #537, opened 2017-10-26
add type property to History objects

e-e-e:
This is a proposal to add a type property to history objects.

We have a need in our app to differentiate between memoryHistory and hashHistory, and thought it would generally be useful to have a property on the history objects which makes it easy to identify which type of history is in use.

What are your thoughts?


mjackson:
One of the points of this library is to abstract away the differences when dealing with browser and hash histories so they are insignificant. Seems like if you need to know the `type` then we're not really doing our job. Can you please say more about why you need this?

PR #538, opened 2017-11-03
Prevent error when null is passed to push() or replace()

### LINKED ISSUE #510: Cannot read property 'state' of null
image

Saw this and it's reporting [this line](https://github.com/ReactTraining/history/blob/master/modules/createBrowserHistory.js?utf8=%E2%9C%93#L198) as the offender. Simple fix would be to change the line to

```jsx
typeof path === 'object' && path !== null && path.state !== undefined && state !== undefined
```

Thoughts?

sman591:
Because `typeof null` evaluates to "object", it's possible that an error will be thrown when attempting to access `path.state` when the path is null. Adding an additional check for null ensures path is always an object before continuing through the conditional.

Fixes #510

Happy to add a test for this as well, but being that `null` isn't a normal value to pass in I wasn't sure if this warrants a test.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,7 @@
+## HEAD
+
+- **Bugfix:** Fix error thrown when passing `null` to `history.push` or `history.replace` ([#538])
+
 ## [v4.6.3]
 > Jun 20, 2017
 
--- a/modules/createBrowserHistory.js
+++ b/modules/createBrowserHistory.js
@@ -153,7 +153,7 @@ const createBrowserHistory = (props = {}) => {
 
   const push = (path, state) => {
     warning(
-      !(typeof path === 'object' && path.state !== undefined && state !== undefined),
+      !(typeof path === 'object' && path !== null && path.state !== undefined && state !== undefined),
       'You should avoid providing a 2nd state argument to push when the 1st ' +
       'argument is a location-like object that already has state; it is ignored'
     )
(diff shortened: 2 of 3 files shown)


mjackson:
`null` isn't a valid argument to `push` or `replace`. You can't push `null`. Can you please say more about why you need this?

sman591:
@mjackson I understand `null` isn't a valid argument to `push` or `replace`, but that doesn't necessarily mean it won't accidentally be passed in by someone's code. It's very easy for a `null` (or other unexpected value - but null is naturally common because it is part of the language) to sneak its way into code paths when things go wrong.

What I'm more proposing here is a small fix to **prevent misleading errors**.

The only reason `null` currently raises an exception here is because this "warning" feature insufficiently checks the `path` variable's value. If the "warning" feature didn't exist, this wouldn't be an issue.

```
!(typeof path === 'object' && path.state !== undefined && state !== undefined)
```

The purpose of this code is clearly to check if a value is (a) provided, or (b) an object with a `state` attribute. However, to properly check for this, a check for null needs to be performed before attempting to access `path.state`

```
!(typeof path === 'object' && path !== null && path.state !== undefined && state !== undefined)
```

mjackson:
I'm all for better error messaging. What I don't agree with is making invalid arguments "work" in a way that we don't really define.

If you'd like to provide some better error messaging, I'd suggest you make another PR that uses the `warning` module to warn people about invalid arguments. Then we can strip that code out of production builds.

sman591:
Fair point - will do 👍

PR #541, opened 2017-11-08
Update BrowserProtocol.js

jsruvi:
ability update history state


jsruvi:
ability update history state

PR #544, opened 2017-11-20
hasBasename bugfix

hmate9:
Previously it didn't work when `prefix` had special regex characters (eg `$`)

### DIFF
--- a/modules/PathUtils.js
+++ b/modules/PathUtils.js
@@ -5,7 +5,7 @@ export const stripLeadingSlash = path =>
   path.charAt(0) === "/" ? path.substr(1) : path
 
 export const hasBasename = (path, prefix) =>
-  path.toLowerCase().lastIndexOf(prefix.toLowerCase(), 0) === 0 && "/?#".indexOf(path.charAt(prefix.length)) !== -1
+  path.toLowerCase().indexOf(prefix.toLowerCase()) === 0 && "/?#".indexOf(path.charAt(prefix.length)) !== -1
 
 export const stripBasename = (path, prefix) =>
   hasBasename(path, prefix) ? path.substr(prefix.length) : path
--- a/modules/PathUtils.js
+++ b/modules/PathUtils.js
@@ -5,7 +5,7 @@ export const stripLeadingSlash = path =>
   path.charAt(0) === "/" ? path.substr(1) : path
 
 export const hasBasename = (path, prefix) =>
-  new RegExp("^" + prefix + "(\\/|\\?|#|$)", "i").test(path)
+  path.toLowerCase().lastIndexOf(prefix.toLowerCase(), 0) === 0 && "/?#".indexOf(path.charAt(prefix.length)) !== -1
 
 export const stripBasename = (path, prefix) =>
   hasBasename(path, prefix) ? path.substr(prefix.length) : path
--- a/modules/__tests__/HashHistory-test.js
+++ b/modules/__tests__/HashHistory-test.js
@@ -270,7 +270,7 @@ describeHistory("a hash history", () => {
       expect(history.location.pathname).toEqual("/")
     })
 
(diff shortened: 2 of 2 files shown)


mjackson:
I'm uncomfortable merging this because I think that using a regex in the underlying code is probably too complex for our needs. All we really need to do is make a string comparison that asserts the *next* char is one of the URL-delimiting characters (i.e. `/`, `?`, `#`) or the end of the string.

Would you be open to doing a more extensive refactoring here?

Parakleta:
I use the following successfully:

```js
export const hasBasename = (path, prefix) =>
    path.startsWith(prefix) && "/?#".indexOf(path.charAt(prefix.length)) !== -1;
```

This obviously works in the delimited case, but in the end-of-string case it also works because the `charAt` function returns the empty string if the length is past the end, and `indexOf` always returns `0` when called with the empty string as an argument.

hmate9:
@Parakleta I added your change to the PR, just also converting `path` and `prefix` to lowercase to make sure the matching is case insensitive.

hmate9:
I don't know why GitHub says there is a merge conflict, there isn't. It is a one line change.

hmate9:
@mjackson could you please comment on this PR. Is it ready to be merged?

modules/PathUtils.js
```diff
@@ -5,7 +5,7 @@ export const stripLeadingSlash = path =>
   path.charAt(0) === "/" ? path.substr(1) : path
 
 export const hasBasename = (path, prefix) =>
-  new RegExp("^" + prefix + "(\\/|\\?|#|$)", "i").test(path)
+  path.toLowerCase().lastIndexOf(prefix.toLowerCase(), 0) === 0 && "/?#".indexOf(path.charAt(prefix.length)) !== -1
```
mjackson: Is there a particular reason you're using `lastIndexOf` instead of `indexOf` here?

mjackson:
Thanks for working on this, @hmate9. Looks great. Just curious about the usage of `lastIndexOf` instead of `indexOf`.

Parakleta:
`lastIndexOf` with a `fromIndex` of `0` is functionally equivalent to `startsWith` in this case but is available everywhere (`lastIndexOf` is ES1, `startsWith` is ES6).  Alternatively you could use something like `path.substr(0, prefix.length) === prefix` which is ES3.

hmate9:
@Parakleta Are you sure about that? According to the docs:

> The indexOf() method returns the index within the calling String object of the first occurrence of the specified value.

It wouldn't make sense to look for more than one match if you have already found one.

Parakleta:
You’re assuming you find a match...

hmate9:
@Parakleta cool, makes sense. Thank you. What do you think @mjackson ? Should I change it back to `lastIndexOf`?

hmate9:
@mjackson could we get an update on this please?

hmate9:
ping @mjackson 

This is ready to be merged and is pretty useful fix

avrahams1:
Is this going to be merged soon?

bjarkehs:
Any news on this? I just ran into this issue myself.

ajhyndman:
I'm running into this too.  I also realized that there are 4 other PRs that have been opened since, attempting to fix the same bug.

#608 
#607 
#604 
#566

marconi:
we just ran into this as well LGTM! 👍

mjackson:
Thanks for the PR, @hmate9, and sorry for sitting on it for so long. I just merged this manually. It will be released in 4.10 and will be present in the 5.0 release as well.

Parakleta:
I think you should be using one of `substr`, `startsWith`, or `lastIndexOf` as discussed in https://github.com/ReactTraining/history/pull/544#pullrequestreview-78387774 in place of the `indexOf(prefix.toLowerCase())`.

Issue #545, opened 2017-11-22
Support asynchronous blocking


kirill-konshin:
It's really hard to implement asynchronous procedure that should delay navigation.

It would be great if blocker function could return `Promise` (if it's resolved navigation should proceed, if rejected — stopped) or a `function` that can accept an argument callback so that `callback(true)` will resume nav.

In this case `result` in https://github.com/ReactTraining/history/blob/master/modules/createTransitionManager.js#L26 treatment should be a bit different.

Change is super small and should not break anything. I propose 2 options (I have made a pull request #546 with scenario 2 because it's more user friendly):

## 1. Less changes: use `getUserConfirmation` to handle string and any other object results

```js
  var confirmTransitionTo = function confirmTransitionTo(location, action, getUserConfirmation, callback) {
      console.log('Bar', prompt, getUserConfirmation);
    // TODO: If another transition starts while we're still confirming
    // the previous one, we may end up in a weird state. Figure out the
    // best way to handle this.
    if (prompt != null) {
      var result = typeof prompt === 'function' ? prompt(location, action) : prompt;

// NEW STUFF
      if (typeof result === 'string' || typeof result === 'object' || typeof result === 'function') {
// NEW STUFF
        if (typeof getUserConfirmation === 'function') {
          getUserConfirmation(result, callback);
        } else {
          warning(false, 'A history needs a getUserConfirmation function in order to use a prompt message');

          callback(true);
        }
      } else {
        // Return false from a transition hook to cancel the transition.
        callback(result !== false);
      }
    } else {
      callback(true);
    }
  };
```

## 2. More user friendly: handle `Promise` right there in transition confirmation

```js
  var confirmTransitionTo = function confirmTransitionTo(location, action, getUserConfirmation, callback) {
      console.log('Bar', prompt, getUserConfirmation);
    // TODO: If another transition starts while we're still confirming
    // the previous one, we may end up in a weird state. Figure out the
    // best way to handle this.
    if (prompt != null) {
      var result = typeof prompt === 'function' ? prompt(location, action) : prompt;

      if (typeof result === 'string') {
        if (typeof getUserConfirmation === 'function') {
          getUserConfirmation(result, callback);
        } else {
          warning(false, 'A history needs a getUserConfirmation function in order to use a prompt message');

          callback(true);
        }
// NEW STUFF
      } else if (typeof result === 'function') {
        result(callback);
      } else if ('then' in result && 'catch' in result) { // duck typing here for simplicity
        result
[...]

PR #546, opened 2017-11-22
Fix for #545

kirill-konshin:
I've implemented the more user-friendly scenario.


mjackson:
This fix won't work for the `beforeunload` event. Please see [my comment here](https://github.com/ReactTraining/history/issues/545#issuecomment-357323951).

PR #548, opened 2017-11-30
Fix browser url does not update while app launched in iframe

n-srg:
In case application launched in iframe, hashHistory modifies iframe url
and Browser address does not get updated because we use window.location
to handle url instead of window.top.location. This commit add new
constructor parameter bypassIframeLocation which forces router to
use window.top.location

This is an enhancement commit, does not replace any current functionality

### DIFF
--- a/modules/createHashHistory.js
+++ b/modules/createHashHistory.js
@@ -34,22 +34,29 @@ const HashPathCoders = {
   }
 }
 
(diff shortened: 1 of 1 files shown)


n-srg:
hey @mjackson could you please have a look on it? I was wondering to get some cool conversation related to this issue :)

PR #549, opened 2017-11-30
add/expose history and location type checkers

lutangar:
Hi there,

I'm proposing to add `history` and `location` type checkers to this project.

At first it might looks like these `prop-types` don't belong here, but here is a few reason why it should:
1. This is the project where functions and properties are defined for history objects.

2. Per the `prop-types` README, it is not React only: 
> You can use prop-types to document the intended types of properties passed to components. React (and potentially other libraries—see the checkPropTypes() reference below) 

3. You don't have to use it, that's completely optionnal.

4. It might help to have tests checking if all properties are always defined across **history** implementation and lifecycle; For example, I'm not sure if it was intended that the `entries`, `index` and the location `key` properties could be `undefined`.

Also, I tried to define a maximum number of properties as `required`, and either way, I'd like to ask if it's always the case or of I am missing some edge cases.

Cheers.


mjackson:
This is out of scope for this project.

PR #555, opened 2017-12-28
Fix history.listen triggered by initial page load on some old browsers

yangmingshan:
Popstate event will fire after initial page load on some old browsers, like Safari on iOS 9. It will trigger history.listen.

This PR is a hack to fix this problem, the solution is from [page.js](https://github.com/visionmedia/page.js). I tested on iOS 9, it worked.

### DIFF
--- a/modules/DOMUtils.js
+++ b/modules/DOMUtils.js
@@ -42,9 +42,17 @@ export const supportsGoWithoutReloadUsingHash = () =>
   window.navigator.userAgent.indexOf("Firefox") === -1;
 
 /**
- * Returns true if a given popstate event is an extraneous WebKit event.
+ * Returns true if a given popstate event is an extraneous WebKit event, or it fired on initial page load.
  * Accounts for the fact that Chrome on iOS fires real popstate events
  * containing undefined state when pressing the back button.
  */
-export const isExtraneousPopstateEvent = event =>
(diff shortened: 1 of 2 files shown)


yangmingshan:
I don't know why the Travis CI always timeout 😔

mjackson:
Thanks for the PR, @yangmingshan. Is this code lifted directly from page.js? If so, I'd like to include a comment that tells people where we got it so we give attribution where it's due.

Would you mind rebasing on top of the current master?

yangmingshan:
Rebased. And yes, this code is from page.js, you can find in [here](https://github.com/visionmedia/page.js/blob/master/index.js#L576-L600).

How I get the testing pass, is there something wrong？

yangmingshan:
The solution source information added.

danywheeler:
What's the status here? today I run into the same problem.

yangmingshan:
@mjackson Can you look at it? It's been a long time.

mjackson:
Sorry for the delay here. This will be released in 4.10.

mjackson:
Hi @yangmingshan, I merged this but it causes one of our tests to fail on Firefox. You can see the failure here: https://travis-ci.org/ReactTraining/history/jobs/583498880

I'm going to revert the commit for now, but I'd be open to trying again if you can figure out what's going wrong in Firefox. Thanks!

PR #561, opened 2018-01-21
Fix eslint errors

guidobouman:
These broke when Prettier was enabled.

I've also added some improvements to the Travis CI setup to prevent this form happening again.


mjackson:
Looks good, thanks @guidobouman!

guidobouman:
💃

PR #568, opened 2018-02-09
Adding onPush and onBlock callbacks to the push methods

nmalzieu:
Proposition : ability to add callbacks to the push method, one if the transition was applied and one if it was blocked.

It enables the use of `history.block` and a one-time callback if the transition worked (I don't think it would be easy to do it with a transition listener?)

If I'm wrong or this is anti-pattern I would be pleased to hear an alternative :)


akashshinde:
Do we have any plan to merge this PR ?

If not then is there any mechanism to call callback function which will get executed once `push()` is done ?

PR #570, opened 2018-02-12
Link method which prevent same locations in the history

dabit1:
I created a method that call `push` when current location and next location is different and call `replace` when they are the same.

It fixes a known problem which was reported for many users. 
[https://github.com/ReactTraining/history/issues/470](https://github.com/ReactTraining/history/issues/470)
[https://github.com/ReactTraining/history/pull/558](https://github.com/ReactTraining/history/pull/558)
[https://github.com/ReactTraining/history/issues/507](https://github.com/ReactTraining/history/issues/507)


modules/createBrowserHistory.js
```diff
@@ -308,6 +308,19 @@ const createBrowserHistory = (props = {}) => {
     };
   };
 
+  const link = (path, state) => {
+    const currentLocation = {
```
guidobouman: You're replication this code in each history implementation. Should it not be part of the `shouldReplace` method?

CHANGES.md
```diff
@@ -1,3 +1,10 @@
+## [v4.6.4]
```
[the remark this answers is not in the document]
dabit1: Yes you're right!

modules/createBrowserHistory.js
dabit1: Do you mean that `shouldReplace` should also call `push`/`replace`?

guidobouman: No,I mean that you can move the following lines to the `shouldReplace` method:

```js
    const currentLocation = {
      path: history.location.pathname + history.location.search + history.location.hash,
      state: history.state || {}
    };
```

modules/createHashHistory.js
```diff
@@ -327,6 +327,19 @@ const createHashHistory = (props = {}) => {
     };
   };
 
+  const link = (path, state) => {
+    const currentLocation = {
+      path: history.location.pathname + history.location.search + history.location.hash,
+      state: history.state || {}
+    };
+
+    if (shouldReplace(currentLocation, {path})) {
```
guidobouman: Should `state` not also be passed to this method, as it's in the params?

modules/createBrowserHistory.js
dabit1: I think not because depends on the history implementation you get this data from history or window.

modules/createHashHistory.js
dabit1: There is no state in hash history

guidobouman: Then remove it from the method params & default data?

modules/createBrowserHistory.js
guidobouman: Then you could pass the history object to the method?

dabit1: Ey @guidobouman  take a look at refactors I did, thanks!

modules/LocationUtils.js
```diff
@@ -74,3 +74,7 @@ export const locationsAreEqual = (a, b) =>
   a.hash === b.hash &&
   a.key === b.key &&
   valueEqual(a.state, b.state);
+
+export const shouldReplace = (location, nextPath, history, nextState) =>
```
guidobouman: I find this function a bit hard to read. Especially with the mixed parameters.

Just thinking out loud here, I could be very wrong, but what if you joined the params?

```js
shouldReplace(currentLocationAndState, newPathAndState) { ... }

...

if(shouldReplace(
  { location: history.location, state: history.state },
  { path, state }
)) {
  replace(path, state);
} else {
  push(path, state)
}
```

README.md
```diff
@@ -138,8 +138,10 @@ The `action` is one of `PUSH`, `REPLACE`, or `POP` depending on how the user got
 * `history.goBack()`
 * `history.goForward()`
 * `history.canGo(n)` (only in `createMemoryHistory`)
+* `history.link(path, [state])`
```
guidobouman: I think `push` & `replace` should go down, or `link` should go up, no?

TrentWDB:
Is there a workaround for this issue? It seems like this has been an issue for the last ~3 years.

dabit1:
@TrentWDB I did not have much time lately. In max 1week I hope to review it ;)

guidobouman:
Hey @mjackson, I think @dabit1 has a great PR here.

It fixes the long standing issue with links & history.push in a solid way. Which would in turn allow for a small PR in react-router to finally let the Link component start behaving like a true link. 🎉

README.md
dabit1: Go up I think is right. Done.

modules/LocationUtils.js
dabit1: I think `link` method is easy to understand now. Not the same with `shouldReplace` method, so I have made some changes in order to be easier to understand it. Tell me what you think.

README.md
```diff
@@ -138,8 +138,10 @@ The `action` is one of `PUSH`, `REPLACE`, or `POP` depending on how the user got
...
 
 When using `push` or `replace` you can either specify both the URL path and state as separate arguments or include everything in a single location-like object as the first argument.
+You also could need to do push only when the current location is different from the next location and do replace when they are the same. To do that you have `link` method.
```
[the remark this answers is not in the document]
dabit1: I agree. Done!

nksfrank:
What's the status on this PR?

guidobouman:
Waiting for the repo owner @mjackson to merge this, after that a new PR for react-router can be made.

sabbiu:
I am also looking for this fix. What is the status?

darklightblue:
@mjackson This type of fix seems to have been discussed ad nauseam for an incredibly long time; even a couple of PRs have vanished. Would be great to get this fixed, finally.

CHANGES.md
```diff
@@ -1,3 +1,10 @@
+## [v4.7.0]
```
mjackson: Please don't add a version number or publish date in your PR (see [our contributing guidelines](https://github.com/ReactTraining/history/blob/master/CONTRIBUTING.md)).

modules/LocationUtils.js
```diff
@@ -74,3 +74,10 @@ export const locationsAreEqual = (a, b) =>
   a.hash === b.hash &&
   a.key === b.key &&
   valueEqual(a.state, b.state);
+
+export const shouldReplace = (location, newPath, history, newState) => {
+  const currentPath = location.pathname + location.search + location.hash;
+  const hasState = history && history.state;
+
+  return currentPath === newPath && !hasState || valueEqual(history.state, newState);
```
mjackson: `history.state` isn't part of our API. Maybe you meant `location.state`?

I believe a test would have caught this mistake. Would you mind including one with this PR? The test should make a few assertions:

1. When we use `history.link` to navigate to the same path, a new entry is not pushed onto the stack...
2. ...unless the state has changed.

jer8:
Please can we have this PR updated @dabit1

infalse:
Hello. Struggled with the same problem of react-router and react-router-dom Link's being PUSHing history even if i was on the same /route already. What worked for me was: 

```
import { withRouter } from 'react-router' 
import { NavLink } from 'react-router-dom' 
import { default as BSNavLink } from 'reactstrap/lib/NavLink' 

export const StyledLink = styled(BSNavLink)...

const NavLinkLogged = ({ to, history, staticContext, ...rest }) => {
  const shouldUseReplace = history.location.pathname === to
  return (
    <StyledLink to={to} tag={NavLink} replace={shouldUseReplace} {...rest} />
  )
}

export default withRouter(NavLinkLogged)
```

replace attribute becoming true when current location matches "to" path. So first time you click will be PUSH, and the others REPLACE.

Hope is helps

undrafted:
I solved this problem by checking 'match' instead:

` {match && match.isExact
            ? <a className={clsNavLink}>{content}</a>
            : <NavLink to={url} className={clsNavLink}>
              {content}
            </NavLink>}`

CHANGES.md
dabit1: done

modules/LocationUtils.js
dabit1: You were right. I created 4 tests to ensure that `shouldReplace` function works properly.

dabit1:
Hi @mjackson! I did some changes in order to have it ready to be merged. A lot of people is requesting it so if you agree I think it's a good improvment.

modules/__tests__/LocationUtils-test.js
```diff
@@ -160,3 +160,71 @@ describe("createLocation", () => {
     });
   });
 });
+
+describe("shouldReplace", () => {
+  it("should return true if current path is the same than new path", () => {
```
guidobouman: Engilsh: `same as` or `other than`, `than` is exclusive, where `as` is inclusive. This is repeated in each test description.

modules/__tests__/LocationUtils-test.js
```diff
@@ -160,3 +160,71 @@ describe("createLocation", () => {
...
+      'the/path?the=query#the-hash',
+      {
+        foo: 'bar',
+        obj: {
+          foo2: 'var3'
+        }
+      }
+    )).toBeFalsy();
+  });
+});
```
guidobouman: I think there's supposed to be a newline at the end of each file.

modules/__tests__/LocationUtils-test.js
dabit1: Ok, done. Thanks.

modules/__tests__/LocationUtils-test.js
dabit1: Done, anyway linter does not have this rule. If you prefer it would be fine if it has it.

mjackson:
Hey @dabit1, sorry for the delay here. Are you using React Router? If so, would you be open to fixing it there instead of here?

This library is designed to be a minimal abstraction over the browser's built-in history API. Someday we hopefully won't need it anymore. That's the only reason I'm hesitant to add more API that the browser has no equivalent for.

mjackson:
Hey @dabit1, sorry for the delay here. Are you using React Router? If so, would you be open to fixing it there instead of here?

This library is designed to be a minimal abstraction over the browser's built-in history API. Someday we hopefully won't need it anymore. That's the only reason I'm hesitant to add more API that the browser has no equivalent for.

Having said that, I'm ready to move on this now. This bug is super annoying and has been here way too long. So whether we fix it here or in the router, it'll be fixed in our next release.

CHANGES.md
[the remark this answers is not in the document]
mjackson: Don't update the version number in CHANGES.md. Instead, just use `## [HEAD]` and we'll update it when we cut a release.

dabit1:
Hi @mjackson ! Nice to meet you finally hehe. So yes, I'm using React Router. But the thing is that in some cases I prefer or I have to use the method `push` of `history` instead of the `Link` component from React Router. For example, if you follow the *Container pattern* in React every callback of every presentational component is controlled by a container so presentational components can't decide what they can do after clicking them. For that reason I need to use the method `push` from the container and this is the reason *Link* can't be used.

Anyway I understand your point since this package is not only to be used for React. But this method is useful I think. So... it is in your hands :)

CHANGES.md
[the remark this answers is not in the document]
dabit1: done

tomasklim:
@mjackson hello, when this feature will be ready to use?

siddhant91:
@mjackson  Hello , is there any update on when will this feature be published and ready to be used ?

cocacrave:
ETA please.

vladshcherbin:
@cocacrave close to never since this PR is more than 1 year old. The long-standing issue will be 2 years soon. 👏

cocacrave:
Anybody have a workaround you recommend and use?

tomasklim:
Somewhere I found this and it works fine:
`
let lastLocation = null;

history.listen(location => {
  lastLocation = location
});
const prevHistoryPush = history.push;

history.push = (pathname, state = {}) => {
  if (lastLocation === null || pathname !== lastLocation.pathname + lastLocation.search + lastLocation.hash ||
    JSON.stringify(state) !== JSON.stringify(lastLocation.state)) {
    prevHistoryPush(pathname, state);
  }
};
`

cocacrave:
@tomasklim Thanks!

vladshcherbin:
Jesus, after all years, issues, PRs and promises the maintainer just closes the PR with no comments like nothing ever happened.

Man, I wish there was another package/router so I would never ever use anything from you.

JUST WOW

🤦‍♂️

guidobouman:
Hey @mjackson,

If you don't have the time or drive to maintain this package, that's fine.

But might I suggest to let a team of contributors (that are willing to make sure this project stays alive) take care of it then? You could still be the owner, but others could help you out and implement the vision for the project. To shorten the time needed to get a PR merged, or get answers on issues.

fabe:
Here's another workaround I used a while ago to get around this (in my case, a page transition was called when navigating to the same page). Warning, it's hacky:

```js
const getUserConfirmation = (location, callback) => {
  const [newUrl, action] = location.split('|');

  // Check if user wants to navigate to the same page.
  // If so, we don't want to trigger the page transitions.
  // We have to check the `action`, because the pathnames
  // are the same when going back in history 🤷‍
  const currentUrl = `${window.location.pathname}${window.location.search}`;
  if (newUrl === currentUrl && action === 'PUSH') {
    callback(false);
    return;
  }

  callback(true);
};

// ...

let history = createHistory({ getUserConfirmation });

// `block` must return a string to conform.
// We send both the pathname and action to `getUserConfirmation`.
history.block((location, action) => {
  return `${location.pathname}${location.search}|${action}`;
});
```

vladshcherbin:
@guidobouman there is now a roadmap in issues for v4 and 5. However, this issue is not in the list, so I guess we are left with dirty hacks forever. 🤷‍♂️

vladshcherbin:
For all watchers, there is another try now - https://github.com/ReactTraining/history/issues/689#issuecomment-478152510 🚀

dabit1:
It's little bit annoying that I worked on it and then it's closed with no clear reason.

mjackson:
Hey everyone, I just realized that I never commented on this issue when I closed it. Sorry about that!

For anyone who is wondering, I left [an explanation in another issue](https://github.com/ReactTraining/history/issues/827#issuecomment-682331904) about why I decided to leave this functionality out of the history library.

It basically boils down to this: `history.push` is just a substitute for `window.history.pushState`. It shouldn't try to be "smart" about doing e.g. a push vs. replace. If it did, we would lose something that the `pushState` API actually gives us, namely the ability to push the same URL onto the stack twice. There are valid use cases for doing so, and overloading the meaning of "push" in this library would just cause more issues and confusion.

However, the original concern is totally legit. This is why React Router's `<Link>` component is smart about this in v6 (currently in beta). If you take a look at the source, you'll notice that [a `<Link>` uses a replace instead of a push when the URL is the same](https://github.com/ReactTraining/react-router/blob/761f104ed144601b3bcd7a8a35aa635afc5e05db/packages/react-router-dom/index.tsx#L203-L206). I think this is the right place for this logic.

This way, history + React Router have parity with the DOM. Our `history.push` works exactly like `window.history.pushState` and React Router's `<Link>` works like a normal `<a>` element.

Again, my apologies for not following up here earlier! I definitely should have.

PR #579, opened 2018-04-01
Add params function to browser & memory history

jsakas:
Hello! Recently I have been finding it useful to extend the history object with a params function, so I can easily retrieve the current URL params or a single param. I thought I would add it into the library itself to see if you guys think its useful. 

#### Summary

This feature allows developers to easily access URL params on the current location by calling `history.params(‘foo’)` or `history.params()`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,12 @@
+## [v4.6.4]
+> Jun 20, 2017
+
+- Add params function to browser & memory history
+
+This feature allows developers to easily access URL params on the current location by calling `history.params(‘foo’)` or `history.params()`. 
+
+[v4.6.4]: https://github.com/ReactTraining/history/compare/v4.6.3...v4.6.4
+
 ## [v4.6.3]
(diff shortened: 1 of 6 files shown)


pshrmn:
Hey @jsakas, thanks for the PR. However, I don't think it is likely that this functionality will be merged because:

1. It is opinionated. `history` doesn't care about how query strings are formatted because it just mimics the [Location API](https://developer.mozilla.org/en-US/docs/Web/API/Location), which has a `search` string. The actual parsing is left up to the user. While `qs` is the most popular query string parser, there are many other options with different functionalities that some people may prefer, so `qs` might not work for everybody.
2. It doesn't need to be a part of `history`. The only part of the code that is `history` related is that it grabs `history.location` to parse the `search` string.

I don't make decisions about what gets merged, so if you have a strong argument for why this should exist, _maybe_ it will get added by Michael, but I personally think this would be better to live outside of `history`.

PR #584, opened 2018-04-06
Expose and use action constants

tobiaszwalczak:
Hello! In an application of mine I'm comparing the action of state transitions with plain strings. I think it would be nice if the library exported the action strings as constants so users of the library could use them.

**Summary**
This feature allows developers to import the constants `POP`, `PUSH` and `REPLACE` from the history package. `import { POP, PUSH, REPLACE } from 'history';`

### DIFF
--- a/modules/constants.js
+++ b/modules/constants.js
@@ -1,0 +1,3 @@
+export const POP = "POP";
+export const PUSH = "PUSH";
+export const REPLACE = "REPLACE";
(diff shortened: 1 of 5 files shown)


tobiaszwalczak:
Can someone take a look a the failing test? I don't really know why it's failing

pshrmn:
I'm not sure why the test failed. I restarted the tests and they passed, so maybe a cache issue?

As far as the PR goes, I'm not really sold on its necessity. If `history` was using an enum it would be nice to export the aliases, but here we're just exporting variables with the string value of their name.

PR #597, opened 2018-06-16
Add sideEffects: false to package.json

smably:
More effective tree-shaking with webpack 4! 🎉


timdorr:
Thanks!

petermikitsh:
@timdorr Could we get a new release with this change? Then webpack can tree shake the history module (as it's typically used with react-router).

PR #603, opened 2018-06-28
Fix to work in createHref using fullpath

hyeonjae:


mjackson:
This library does not deal with protocol or domains, only paths.

Issue #564, opened 2018-02-03
Cannot include URL query in basename


goffioul:
I have a use case where my react app can be accessed using a URL that contains query parameters, e.g. `http://example.com/index.html?foo`. I create the history with `createBrowserHistory({basename:'/index.html?foo'})`. This produces warnings and I believe the path matching will not work. The root cause of the problem is that `PathUtils.hasBasename` does not escape special characters in `prefix` to create its regular expression.

mjackson:
The basename does not support query parameters. It's a segment of the pathname.

PR #604, opened 2018-07-03
Escape prefix value in hasBasename

w33ble:
Closes https://github.com/ReactTraining/history/issues/564

- Adds tests for `PathUtils.hasBasename` and `PathUtils.stripBasename`
  - "hasBasename with a search value given as a url with basename returns true" fails
- Escapes these characters when creating RegExp in `hasBasename`: `| \ { } ( ) [ ] ^ $ + * ? .` 
  - "hasBasename with a search value given as a url with basename returns true" now passes

This PR fixes `PathUtils.hasBasename` so that it correctly handles special URL characters. This allows users to specify a basename that contains a search params and other special regex characters and have it both correctly match and correctly strip the basename from the path.

This is helpful when you want to be able to ignore everything in the URL up to a given point. For example, given the URL `/basepath/app?search=params&to=ignore#/real/path/to/parse`, you may only care about `/real/path/to/parse`. Providing a basename value of `/basepath/app?search=params&to=ignore#/` would not previously work, because the `?` was not escaped, so the produced RegExp in `hasBasename` would not match correctly.

This could be considered a breaking change if you were relying on the internal checking mechanism to use a regex pattern instead of a string. However, given that the docs specifically call for the basename value to be a *string*, I'd consider that an incorrect use of the basename value in the first place. From the README (emphasis mine):

> If all the URLs in your app are relative to some other "base" URL, use the basename option. This option transparently adds **the given string** to the front of all URLs you use.

*Aside*: The contributing guide says to add a note to CHANGES.md, but it seems that hasn't been done since `v4.6.3`, so I skipped that step... looks like most recent PRs haven't done this either.

### DIFF
--- a/modules/PathUtils.js
+++ b/modules/PathUtils.js
@@ -7,7 +7,7 @@ export function stripLeadingSlash(path) {
 }
 
 export function hasBasename(path, prefix) {
-  return new RegExp("^" + prefix + "(\\/|\\?|#|$)", "i").test(path);
+  return new RegExp("^" + prefix.replace(/[|\\{}()[\]^$+*?.]/g, '\\$&') + "(\\/|\\?|#|$)", "i").test(path);
 }
 
 export function stripBasename(path, prefix) {
--- a/modules/__tests__/PathUtils-test.js
+++ b/modules/__tests__/PathUtils-test.js
@@ -1,0 +1,78 @@
+import expect from "expect";
+import { hasBasename, stripBasename } from "../PathUtils";
+
+describe("hasBasename", () => {
+  describe("with a simple path", () => {
+    const basepath = '/prefix/with'
+    describe("given as a url with basename", () => {
+      it("returns true", () => {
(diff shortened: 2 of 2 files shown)


a-neumann:
I have a similar problem with the use of [] and | characters in the path which are interpreted unescaped by the regexp.

a-neumann:
@w33ble i created https://github.com/ReactTraining/history/pull/607 that adds some other problematic characters to your fix. Maybe you could add these to your PR so i could close mine. Thanks in advance

w33ble:
@a-neumann I borrowed from #608, which used `escape-string-regexp`. I just grabbed the logic from that module and put it into my PR. The following characters are all escaped here: `| \ { } ( ) [ ] ^ $ + * ? .`

w33ble:
This is failing with the same errors that https://github.com/ReactTraining/history/pull/634 failed with (that PR was still merged).

Tests all run locally, maybe something is wrong with the travis setup?

w33ble:
I suppose it's time to fork. :(

I'll leave the branch around in case someone else wants to try making it happen.

PR #607, opened 2018-07-18
Escape prefix value in hasBasename

a-neumann:
fix: hasBasename handles search and additional special characters correctly

escape the ? [, ] and | character before using it in the regexp instantiation


a-neumann:
@w33ble thanks a lot for creating the issue #604 and writing the tests! I created a PR based on your changes.

a-neumann:
Using `escape-string-regexp` instead will be a better more comprehensive solution.

PR #608, opened 2018-07-19
Escape basename before interpreting as Regex

### LINKED ISSUE #605: Misleading basename warning message
Hello,

This is going to be quite considering the severance of the issue but maybe someone can point me somewhere though I know very well that the root of my main problem has nothing to do with the code in this repository. I just stumbled upon a minor issue during my research.

I am developing a React web application which needs to be called with a specific query string containing a parameter. My application needs to be integrated with a third-party proprietary component which renders a dynamic header and a footer in the application.

This header and footer contain `<a>` tags with the attribute `href="#"`. Of course these links don't go anywhere but they have `onclick `handlers which expand/collapse menus etc. At least that's how they should work. But within the head there is a `<base>` tag which's value is the relative path to the root of the application but without the query string. This tag is rendered on the server and this makes sure that the application works the same way in different environments (e.g. http://example.com and http://example.com/Foo as well without any modifications).

However this gives me a headache right now as the relative anchor tags compute their absolute target urls relative to the base path. That does not include the query string and therefore upon clicking them the browser detects an URL change and performs a redirect instead of just appending the hashmark at the end of the URL.

So I have tried to modify the `<base href>`. If I change it in the inspector and append the query string, everything works fine. But if I change it in the code and try to load the application, nothing renders and there is no error message at all - except for a warning:

> Warning: You are attempting to use a basename on a page whose URL path does not begin with the basename. Expected path "/Foo/?bar=777" to begin with "/Foo/?bar=777".

My issue lies here: the warning message is simply not true, because the strings are identical and therefore they both begin with one another. I have found that the relevant check originally tested this exact condition (`indexOf... !== -1`) described in the warning message, but in a later commit it has been replaced with a more advanced logic with regex matching etc: (https://github.com/ReactTraining/history/commit/e07508777843bc9a0e45e32e69e1d3330ef5973e#diff-f62943cda47e8992518ce1f1e28f38c1)

But the warning message remained unchanged. I think this is misleading because it does not correctly describe the actual problem and states a fact which is false.

ajhyndman:
Fixes #605 

If you'd rather not introduce a new dependency, I'm open to suggestions, but `escape-string-regexp` is literally 11 lines of code.

https://github.com/sindresorhus/escape-string-regexp/blob/master/index.js

### DIFF
--- a/modules/PathUtils.js
+++ b/modules/PathUtils.js
@@ -1,3 +1,5 @@
+import escapeStringRegexp from 'escape-string-regexp';
+
 export const addLeadingSlash = path =>
   path.charAt(0) === "/" ? path : "/" + path;
 
@@ -5,7 +7,7 @@ export const stripLeadingSlash = path =>
   path.charAt(0) === "/" ? path.substr(1) : path;
 
 export const hasBasename = (path, prefix) =>
-  new RegExp("^" + prefix + "(\\/|\\?|#|$)", "i").test(path);
+  new RegExp("^" + escapeStringRegexp(prefix) + "(\\/|\\?|#|$)", "i").test(path);
 
 export const stripBasename = (path, prefix) =>
   hasBasename(path, prefix) ? path.substr(prefix.length) : path;
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
   },
   "dependencies": {
     "babel-runtime": "^6.26.0",
+    "escape-string-regexp": "^1.0.5",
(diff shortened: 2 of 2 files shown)


ajhyndman:
Oh, I just realized that there are actually a half dozen other open PRs that also address this issue.

#607 
#604 
#566 
#544 

#544 might be the most robust, since it avoids ever attempting to parse the basename as a RegExp.

ajhyndman:
I'm going to close this PR in favour of #544!

PR #611, opened 2018-07-26
Allow getUserConfirmation callback result to cater for any type of value

### LINKED ISSUE #489: confirmTransitionTo prompt require type string, maybe it shouldn't?

mymattcarroll:
Closes #489

We'd like to have a multi-option confirmation popup (save/discard/cancel) and need a way to pass additional details about the actions, and an action handler into getUserConfirmation.

The changes proposed will not break break compatibility but will allow for any type of value returned from the callback.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,7 @@
+## HEAD
+
+- Allow `getUserConfirmation` callback result to cater for any type of value ([#489])
+
 ## [v4.6.3]
 > Jun 20, 2017
 
--- a/modules/createTransitionManager.js
+++ b/modules/createTransitionManager.js
@@ -26,24 +26,23 @@ const createTransitionManager = () => {
       const result =
         typeof prompt === "function" ? prompt(location, action) : prompt;
 
-      if (typeof result === "string") {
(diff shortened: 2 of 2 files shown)


mymattcarroll:
@mjackson, any chance of getting this merged? We would really like this feature in https://github.com/FormidableLabs/redux-little-router

mjackson:
Hey @mymattcarroll, thanks for the PR. Any apologies for the delay following up. Can you please describe a bit more about what you're trying to do? The `getUserConfirmation` method is giving a lot of people trouble, so I'm currently on the look-out for a better solution.

mymattcarroll:
@mjackson as this was quite a while ago, I don't remember the exact use case but we wanted to pass a more complex object as the `result` which was more than just JSON so we would not stringify it either. If you need more information, I might have to wait until the weekend to did back into the code base we were using this lib in.

Issue #477, opened 2017-05-18
Allow forceRefresh to be toggled after instantiation.


Munksey:
I think it would be nice if `forceRefresh` were configurable, even after instantiation.

The particular use case that I am envisioning is for gracefully forcing clients to update their versions. We periodically check to see if the current client is outdated and if so, it would be nice to make the next click on a link reload the application.

Another reason I think this would be nice is that `<BrowserRouter>`, in the react-router library, has a flag that can be toggled, but has no impact after instantiation. This makes the component uncontrolled, unfortunately. This would be an argument for making all configurations toggle-able.

I don't know what the API would look like. But, to keep it generic and re-useable across the different APIs, perhaps we want something like `history.setConfig('forceRefresh', true)` or `history.setConfig({ forceRefresh: true })`. 

Thoughts?

mjackson:
`forceRefresh` isn't something that should change over time. Either you need to get a new page from the server when the URL changes or you don't.

Having said that, there are ways to work around this if you need them. For your use case you can just not use `forceRefresh` and use `window.location.reload` when the user clicks your link.

getkey:
It would make sense for my use case: in some pages of our website we let advertising partners inject ads. Basically we load a script from them so they have full access to our website to make changes (I know it sucks but it's how it's traditionally done).

The problem is that we don't know what they might do so I'd like to be able to refresh the page **only** when the user was on a modified page, to make sure the page is properly cleaned up.

PR #614, opened 2018-08-04
Allow changing forceRefresh

stipsan:
# TL;DR https://codesandbox.io/s/n77x71v6x4

This has been discussed in #477. The [suggestion](https://github.com/ReactTraining/history/issues/477#issuecomment-307545239) then were to use `window.location.reload` when the user clicks on a link. Having tested this solution in production I feel this topic is worth a revisit 🤔 

The [codesandbox](https://codesandbox.io/s/n77x71v6x4) demonstrates the difference between using `window.location.reload` and having the option to set `forceRefresh` by calling `history.setForceRefresh` itself in a typical `react-router` setup.

Since the call to `window.location.reload` is async the page usually unloads after rendering the same route that is being navigated to, making it look like something went wrong to the end user:

kapture 2018-08-04 at 19 56 26

Compare it to `forceRefresh`: 

kapture 2018-08-04 at 19 58 22

I tried my best to find alternative solutions but doing something the equivalent of:
```jsx
import {Router} from 'react-router-dom';

const forceRefreshHistory = createBrowserHistory({forceRefresh: true});

...
return <Router history={shouldForceRefresh ? forceRefreshHistory : history}>
```
Turned out to be surprisingly difficult. As was trying to wrap every `<Link />` and `<Redirect />` with custom logic that overrides events and essentially reimplement what `forceRefresh` in `history` does to achieve the same result. But as it turns out there's more logic than react components that might trigger a navigation events so this too is fragile (and a lot of code compared to just rendering a component somewhere wrapped in `withRouter` that calls `history.setForceRefresh(true)` and that's it).

What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -235,6 +235,8 @@ const history = createBrowserHistory({
 })
 ```
 
+It's also possible to change this setting using `history.setForceRefresh(true)`.
+
 ### Modifying the Hash Type in createHashHistory
 
 By default `createHashHistory` uses a leading slash in hash-based URLs. You can use the `hashType` option to use a different hash formatting.
--- a/modules/createBrowserHistory.js
+++ b/modules/createBrowserHistory.js
@@ -42,7 +42,6 @@ const createBrowserHistory = (props = {}) => {
   const needsHashChangeListener = !supportsPopStateOnHashChange();
 
   const {
-    forceRefresh = false,
     getUserConfirmation = getConfirmation,
     keyLength = 6
   } = props;
@@ -150,6 +149,11 @@ const createBrowserHistory = (props = {}) => {
   // Public interface
 
   const createHref = location => basename + createPath(location);
+
+  let {forceRefresh = false} = props;
+  const setForceRefresh = forceNextRefresh => {
+    forceRefresh = forceNextRefresh
+  }
 
   const push = (path, state) => {
     warning(
@@ -319,7 +323,8 @@ const createBrowserHistory = (props = {}) => {
     goBack,
     goForward,
     block,
-    listen
+    listen,
+    setForceRefresh,
   };
 
   return history;


kevinlaw91:
This is also useful for react-router for site force fresh.
e.g. if react app is updated, we can tell history to use force refresh for next navigation

mjackson:
Thank you for the detailed demo, @stipsan! Very nice. I don't think I realized that `window.location.reload()` would wait until the new route was rendered before refreshing the page.

Still, I wonder if `setForceRefresh` is the right API. Changing the configuration of the history object itself is going to change the behavior of all links on the page. But it feels like there could be links on the same page, some of which navigate seamlessly and others which cause a reload.

Maybe it'd be better to expose this behavior in `push`, like `push(url, state, forceRefresh = false)`. At a higher level, we could perhaps provide a `<Refresh path="...">` component or similar in the router that would set this up for you.

wdyt?

mjackson:
Just following up here: this issue has been on my mind for a while now. I'm thinking we can maybe address it in v5 with a new `reload` arg to `push`/`replace` so you can opt-in to full page reloading when you actually initiate navigation instead of when you create your history object. Maybe something like:

```js
history.push(`/gallery`, /* state */ null, /* reload */ true);
```

I have [a `TODO` about it](https://github.com/ReactTraining/history/blob/105b64a99359c4a438abbc8e6735fa277b5640f5/modules/index.js#L275) on [the dev branch](https://github.com/ReactTraining/history/tree/dev).

What do you think, @stipsan? Would that work?

kevinlaw91:
Hi @mjackson, can we have an additional option to switch to global force refresh mode too? This can be useful when server push a "new version available" signal to the web app, and the app will refresh automatically during next navigation.

chaance:

spikebrehm:
Hello folks! Was just looking into the ability to change the `forceRefresh` value and found this.

What we've done in the meantime is extend `History` like so:

```ts
import { Action, History, Location } from 'history';

export interface RefreshingHistory<T> extends History<T> {
  setForceRefresh(val: boolean): void;
}

export default function hardRefreshable<T>(
  h: History<T>
): RefreshingHistory<T> {
  let forceRefresh = false;
  h.listen((loc) => {
    if (forceRefresh) {
      window.location.href = h.createHref(loc);
    }
  });

  return {
    ...h,

    // We need to do a bit more work to proxy the non-method properties
    get length() {
      return h.length;
    },
    set length(l: number) {
      h.length = l;
    },
    get action() {
      return h.action;
    },
    set action(a: Action) {
      h.action = a;
    },
    get location() {
      return h.location;
    },
    set location(l: Location<T>) {
      h.location = l;
    },

    setForceRefresh(val: boolean) {
      forceRefresh = val;
    },
  };
}
```

iamsaumya:
TL;DR - Are there any caveats of implementing this? Do we just have to make the changes in the component rendering router? 

Thanks for posting the workaround 😄 .  I have used this particular snippet in the following way. However, every now and then I stumbled across inconsistencies of the path not being updated correctly. In my app, I've imported `useHistory` hook from `react-router-dom` and use it `push`, `pop` and `goBack` etc.  

```
import * as React from 'react';
import { Router } from 'react-router-dom';
import { createBrowserHistory } from 'history';

import hardRefreshable, { RefreshingHistory } from 'utils/hardRefreshable';

function App() {
  const refreshableHistory: RefreshingHistory<{}> = hardRefreshable(
    createBrowserHistory<{}>()
  );

  React.useEffect(() => {
    const id = setTimeout(
      () => {
        console.log('Hard refresh on');
        refreshableHistory.setForceRefresh(true);
      },
      1000 * 20 // 20 secs for testing purpose
    );
    return () => clearTimeout(id);
  }, [refreshableHistory]);

  return (
    <Router history={refreshableHistory}>
      <RoutesComponent />
    </Router>
  );
}

export default App;
```

PR #617, opened 2018-08-25
Add a link API that navigates without duplicating paths

sximba:
This PR adds a link API that can be used to closely match what browsers do when navigating to the current path, which is to replace instead of push.

I know #570 is already open but I saw no activity on it in a while. I addressed the comments on it.

This fixes a problem that a lot of people have and will allow dependant libraries to close long standing issues.
#470 
#558 
#507

https://github.com/ReactTraining/react-router/issues/6137

### DIFF
--- a/modules/LocationUtils.js
+++ b/modules/LocationUtils.js
@@ -69,6 +69,16 @@ export function createLocation(path, state, key, currentLocation) {
   return location;
 }
 
+export function shouldReplace(location, newPath, newState) {
+  const nextLocation = createLocation(newPath, newState, null, location);
+
+  return (
+    location.pathname === nextLocation.pathname &&
+    location.search === nextLocation.search &&
+    location.hash === nextLocation.hash
+  );
+}
+
 export function locationsAreEqual(a, b) {
   return (
     a.pathname === b.pathname &&
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,8 @@
+## HEAD
+> Nov 02, 2018
+
+- Add `history.link` which navigates and prevents same paths in the history stack 
+
 ## [v4.6.3]
 > Jun 20, 2017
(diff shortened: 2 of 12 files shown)


mjackson:
Thanks for the PR, @sximba. I definitely want to resolve this in our next release, 4.8. Question: are you also using React Router? Or are you using the history library with something else?

sximba:
@mjackson I am using React Router

modules/LocationUtils.js
```diff
@@ -68,6 +68,17 @@ export const createLocation = (path, state, key, currentLocation) => {
...
 };
 
+export const shouldReplace = (location, newPath, newState) => {
+  const nextLocation = createLocation(newPath, newState, null, location);
+
+  return (
+    location.pathname === nextLocation.pathname &&
+    location.search === nextLocation.search &&
+    location.hash === nextLocation.hash &&
+    valueEqual(location.state, nextLocation.state)
```
pshrmn: `state` shouldn't be compared; replicating how an anchor work only depends on the "visible" location properties.

sximba: Fixed

gabsprates:
why this wasn't merged yet?

mjackson:
After thinking about this more, this is definitely a problem I want to solve in the router, not here. A `<Link>` should be smart enough to make the decision to `push` or `replace` w/out adding more API here.

I know it's a little frustrating since ReactTraining/react-router#6137 was already closed and we sent you over here. But thank you for your patience as we sort this out. Let's follow up in https://github.com/ReactTraining/react-router/issues/5362

stefanprobst:
> After thinking about this more, this is definitely a problem I want to solve in the router

could you clarify why not add a way to programmatically do this with `history.link`? thanks!

PR #622, opened 2018-09-24
Encoding changes

wacii:
This includes two changes:

1. Adds configurable decode path function. The library currently uses `decodeURI()` in `createLocation()`, but this only works with strings encoded with `encodeURI()`. Otherwise it can leave the path partially decoded. The configuration option is called `decodePath` and is specified when creating a history object.

2. Passes original encoded path to browser (when `push()`-ing or `replace()`-ing). Currently the decoded path is used, depending on the browser to re-encode the URL. This works, but only if you don't have '%' symbols in your URL. Also, why interfere with the user's encoding here?

There are a couple things yet to do (look at hash history which I'm not using, cleanup commits, update readme and changes), but I thought I'd run this by you all first.

### DIFF
--- a/modules/LocationUtils.js
+++ b/modules/LocationUtils.js
@@ -2,7 +2,7 @@ import resolvePathname from "resolve-pathname";
 import valueEqual from "value-equal";
 import { parsePath } from "./PathUtils";
 
-export const createLocation = (path, state, key, currentLocation) => {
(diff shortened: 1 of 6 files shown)


wacii:
This has been discussed a number of times, but #505 is probably the most current discussion.

PR #623, opened 2018-09-27
Fix logic issue creating duplicate listeners

zerovox:
@timdorr there's a bug in the logic in creating DOM listeners, currently if you attach two listeners and remove one, you'll have 1 listener, so in both browser and hash histories, it'll attach the listeners a second time.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,8 @@
+## HEAD
+- Fix a bug causing DOM listeners to be attached multiple times
+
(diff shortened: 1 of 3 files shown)


zerovox:
Thanks for fixing the underlying issue. I still worry that the only correct arguments to `checkDOMListeners` are 1 or -1 (i.e `if (listenerCount === 1 && delta === 1) {` isn't going to work for anything else), might be worth a comment to that effect.

PR #631, opened 2018-10-19
Add documentation for unregistering listeners

tstirrat15:
## Motivation
We've got some wonky logic in our React app that means that we want to listen for `POP`s. `history` provides this! It's great! But you generally want to clean up listeners when they're no longer needed, e.g. in `componentWillUnmount`. I found this functionality in the source code, but it wasn't documented, so I added documentation.

## Changes
* Add documentation for `unlisten`.


mjackson:
Thanks for the PR, @tstirrat15!

PR #632, opened 2018-10-20
Bundle esm and cjs and remove path imports

TrySound:
Same structure as in react-router

```
esm/
  history.js
cjs/
  history.js
  history.min.js
umd/
  history.js
  history.min.js
```

As you can see the code of this library is almost fully treeshakable.
1kb in webpack is just a runtime. Rollup result is more clean and shows
only one not treeshaked side effect.


TrySound:
@mjackson if you are not ready for major release yet I can try another optimisations.

mjackson:
This looks great, thanks @TrySound. I'd also like to include stubs of all our files that used to be in the root of the build just for the sake of anyone who was using them. Then we can go ahead and ship this in the same version w/out a major version bump.

PR #634, opened 2018-10-29
Replace warning and invariant with tiny versions

TrySound:
`warning` and `invariant` packages are jealously maintained as mirrors
of fbjs versions. Guys do not accept optimizations like esm.

But happily there are alternatives. They are even more lightweight and
save for us 2kB of parsed (minified) code.


TrySound:
@mjackson after this I will submit one more very nice improvement :)

mjackson:
The build is broken. I'm trying to fix it now.

TrySound:
Wait. I found one problem with them. Names are not provided and they are appear in bundle as `function index() {}`. Fixing.

TrySound:
Ref
https://github.com/alexreardon/tiny-invariant/pull/15
https://github.com/alexreardon/tiny-warning/pull/1

TrySound:
Done. Updated to latest versions.

mjackson:
Thanks, @TrySound. This is good stuff. Makes me wanna do the same thing in all my repos!

PR #643, opened 2018-11-28
Clarified docs to explain how to block routes from transitioning

gustavobini:
Added a new comment to docs explaining how to block single routes without showing the window.confirm dialog.


mjackson:
Thank you for the PR, @gustavobini, but I don't think it's necessary to add another example for `history.block` as its primary use is for prompts.

PR #654, opened 2019-01-10
fix: The second param of the  allPaths/allKeys is redundant

helios741:
`prevIndex === -1 ? 0 : prevIndex + 1` and `prevIndex + 1` are completely equivalent, using the latter can also reduce the code

### DIFF
--- a/modules/createBrowserHistory.js
+++ b/modules/createBrowserHistory.js
@@ -183,10 +183,7 @@ function createBrowserHistory(props = {}) {
             window.location.href = href;
           } else {
             const prevIndex = allKeys.indexOf(history.location.key);
-            const nextKeys = allKeys.slice(
-              0,
-              prevIndex === -1 ? 0 : prevIndex + 1
-            );
+            const nextKeys = allKeys.slice( 0,prevIndex + 1 );
 
             nextKeys.push(location.key);
             allKeys = nextKeys;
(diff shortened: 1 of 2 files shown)


modules/createBrowserHistory.js
```diff
@@ -183,10 +183,7 @@ function createBrowserHistory(props = {}) {
             window.location.href = href;
           } else {
             const prevIndex = allKeys.indexOf(history.location.key);
-            const nextKeys = allKeys.slice(
-              0,
-              prevIndex === -1 ? 0 : prevIndex + 1
-            );
+            const nextKeys = allKeys.slice( 0,prevIndex + 1 );
```
guidobouman: `(0, prevIndex + 1);`

modules/createHashHistory.js
```diff
@@ -210,10 +210,7 @@ function createHashHistory(props = {}) {
           pushHashPath(encodedPath);
 
           const prevIndex = allPaths.lastIndexOf(createPath(history.location));
-          const nextPaths = allPaths.slice(
-            0,
-            prevIndex === -1 ? 0 : prevIndex + 1
-          );
+          const nextPaths = allPaths.slice( 0,  prevIndex + 1 );
```
guidobouman: See previous comment.

helios741: If there is a new version of history, isn't this PR unnecessary?

pshrmn:
Closing in favor of #693.

PR #656, opened 2019-01-10
Remove location pathname decoding

OliverJAsh:
Re. https://github.com/ReactTraining/history/issues/505#issuecomment-453175833

Fixes https://github.com/ReactTraining/history/issues/505

To do:

- [x] update tests (I probably need help here)


kevinkyyro:
It looks like a lot of the failing tests are for the behavior you're removing, so they should either be removed as well or updated to assert that the pathname remains encoded (a simple find-and-replace of "/歴史" to "/%E6%AD%B4%E5%8F%B2" might do it)

OliverJAsh:
@kevinkyyro Done! I had to delete some tests which weren't valid anymore.

modules/__tests__/BrowserHistory-test.js
```diff
@@ -64,17 +64,11 @@ describe('a browser history', () => {
...
         TestSequences.PushEncodedLocation(history, done);
       });
     });
 
-    describe('push with an invalid path string (bad percent-encoding)', () => {
-      it('throws an error', done => {
-        TestSequences.PushInvalidPathname(history, done);
-      });
-    });
-
```
OliverJAsh: This test was no longer valid, since we don't throw an error now we have no decoding. (We assume the pathname provided will be correctly encoded.)

OliverJAsh:
@mjackson @kevinkyyro Any updates on this?

oskarolsson-jimdo:
Is there something one could potentially help out with to get this merged?

Quite keen to get these changes merged since it's a rather big issue in some cases.

modules/LocationUtils.js
```diff
@@ -32,21 +32,6 @@ export function createLocation(path, state, key, currentLocation) {
       location.state = state;
   }
 
-  try {
-    location.pathname = decodeURI(location.pathname);
```
kevinkyyro: I wonder if the owners of this library would prefer a warning:
```js
// in try-catch, no sense throwing any longer
if (location.pathname !== decodeURI(location.pathname)) {
  // warn: pathnames are no longer automatically decoded
}
```

OliverJAsh:
Thanks for the review @kevinkyyro. I'm happy to make those changes once I've heard back from a maintainer of this repository, as then I will be able to batch it with any other requested changes. /cc @mjackson @pshrmn

saravanansarz:
@OliverJAsh  when is this expected to **merge** to mainstream branch and released officially or let us know a **workaround** for this, something like overriding **createLocation** function inside our project.

OliverJAsh:
@saravanansarz I am not a maintainer of this repo. We are awaiting feedback from a maintainer.

saravanansarz:
@mjackson @pshrmn  Any updates of merging @OliverJAsh changes ?

OliverJAsh:
@pshrmn I expected this would require a major version bump, but it's got to happen at some point, and there's a lot of demand for it.

What are the next steps to getting this PR merged?

In the meantime, [I commented here](https://github.com/ReactTraining/history/issues/505#issuecomment-462328076) about how we're maintaining a local fork (via patch-package), and how others can do the same. Alternatively, someone could publish a fork online.

pshrmn:
To be frank, I don't know, which is why I suggest publishing a fork. That would also give some indication as to how much demand there is for this.

kevinkyyro:
> ... That would also give some indication as to how much demand there is for this.

Please note that the level of demand can already be seen in #505, which at present is the issue with most 👍 reactions and comments of all issues, open or closed.

oskarolsson-jimdo:
I fail to see why this should be a fork. If it was really only done as a convenience to RR (as indicated by #505), then really it should be done on RR's side and not in this library -- which would make it a lot simpler to turn off by introducing a config flag.

pshrmn:
I've come around a bit on this and while encoded strings will be more annoying to work with, they're the only way that we can correctly support all valid pathnames.

@OliverJAsh can you add a dependency on https://www.npmjs.com/package/encodeurl and wrap the `pathname` in a call from that?
```js
location.pathname = encodeURL(location.pathname);
return location;
```
That would ensure that every `pathname` is fully encoded, whether it comes from the browser or the user.

The `LocationPathnameAlwaysSame` test should also be updated to be "always encoded".

I don't have the power to merge this myself (well, technically I can but the decision isn't really mine to make), but those changes would be enough for me to support putting them out as a v5 release.

I have some thoughts on a future implementation that could support decoding, but they would also be breaking changes that would need to wait for a new React Router version, so that can wait.

OliverJAsh:
Thanks for the update @pshrmn.

Initially I thought this change should disable decoding, instead just using the pathname the user provides. (We've been using this change locally for many months at Unsplash and had no problems.)

However I see you're suggesting that we go further than this and proactively encode the URL (via [`encodeurl`](https://github.com/pillarjs/encodeurl) which only encodes non-encoded segments).

I think it's hard to think through the implications of this.

Is there any reason why we can't we just leave the `pathname` as it is (i.e. let the user decide how/when/if to encode it)?

One other thought I've just had: once this is done, we'll have to remember to decode params in React Router (https://github.com/ReactTraining/history/issues/505#issuecomment-375646353).

pshrmn:
@OliverJAsh Consistency. I believe that the `pathname` should always be the same, no matter if it comes from the browser (where it would be encoded) or the user (where it may be non-encoded, encoded, or partially encoded).

I am open to removing the change if there is a good reason to support mixed `pathname` states, but I feel that that would only lead to bugs.

As far as React Router goes, I think that the "safest" approach would be to release this under `v5`, leave React Router's `history` dependency on `v4`, and clearly document that user's can manually install `history` `v5` to have encoded `pathname`s. This isn't perfect; if there are bug fixes it would mean maintaining two versions. It would also mean that users start out with the "incorrect" implementation, but it is "safer" than bumping everyone up to `v5` and telling everyone whose apps break to downgrade to `v4.

That would give us time to figure out the best approach for React Router v6, which would have encoded `pathname`s as default (although possibly giving the user the opportunity to provided a function for decoding it).

OliverJAsh:
Let's put it to the users: https://github.com/ReactTraining/history/issues/505.

audunhalland:
I've been following this discussion for a while; our project too has problems with history and decoding. A clean and clear API is very important. In my experience, correctness of URL encoding is not something that the average developer pays a lot of attention to. The number of bugs on the internet related to this mess is immense, it's not just react.

"Auto-encoding" of a URL cannot be done unambiguously. Therefore, I think the new major version of the lib should only accept valid URLs as `location`, and IMO could even _throw_ on invalid input, at least never touch or modify it. This would force developers to think more actively about encoding.

Just my 2 cents, as someone who has spent too many hours fixing sloppy URL code (disclaimer: my native alphabet is non-ASCII 😄  )

Dalzhim:
@pshrmn : It seems to me that the best path forward is to apply @kevinkyyro 's suggestion to generate a warning. However, I believe the correct implementation for that warning would be to use the following comparison : `location.pathname !== encodeurl(location.pathname)`. The `encodeurl` function doesn't try to encode anything that is already valid within an URL, thus if users of the library submit well formed URLs, the call to encodeurl should result in no changes at all. Any change means the original url was not well formed and it is a programming mistake that must be solved by the user of the library.

Dalzhim:
@pshrmn : One more thing, considering the timeline to fix this 2year-old known issue is still stretching out, I would urge the maintainers of this repository to accept #675 in order to document this issue appropriately.

jacobwgillespie:
> "Auto-encoding" of a URL cannot be done unambiguously. Therefore, I think the new major version of the lib should only accept valid URLs as `location`, and IMO could even _throw_ on invalid input, at least never touch or modify it. This would force developers to think more actively about encoding.

Just want to echo this - as someone handling URLs with unicode characters, I would much prefer the `history` library behavior to match the browser's behavior.  There are some complex URL scenarios, and IMO it's important that code using `history` is able to have full access to the Location object and have the ability to implement whatever logic is necessary for the specific situation.  

It's also important to avoid divergent URL behavior based on the entrypoint to a specific location, for instance visiting a URL directly, vs clicking on an `<a>` tag, vs navigating back/forward, vs navigating with `pushState`, vs navigating via `history` / `<Link>`.  Those can produce different pathname behaviors if `history` doesn't align with the browser.

Today, that's not possible due to the auto-encoding behavior, so we have to work around this by directly accessing `window.location` and bypassing `history` and the router.

pshrmn:
Throwing on invalid pathnames is an interesting idea.
```jsx
<Link to={`/user/${encodeURIComponent("Beyoncé")}`}>Beyoncé</Link>
```
It is certainly more verbose for users, but if it is safer and has the same end result, I can support that.

I realized that we'll actually have to modify `encodeurl` because it leaves square brackets un-encoded (and undoes their encoding). This is done for IPv6 in a URL's `host` component, but shouldn't happen for the `pathname`.

modules/LocationUtils.js
```diff
@@ -65,6 +52,11 @@ export function createLocation(path, state, key, currentLocation) {
       location.pathname = '/';
     }
   }
+  
+  tinywarning(
+    location.pathname === encodeurl(location.pathname),
+    "location.pathname is not fully encoded, which may result in bugs."
+  );
```
OliverJAsh: I have mixed feelings about libraries directly invoking the console API. In our application, we require all logs to be in JSON format (via a library such as Bunyan/Winston/Pino), so they can be easily queried at production scale.

I'm not sure what else we could do here though—other than provide a way to change the logging effect, defaulting to the native console API.

pshrmn: `tinywarning` is a no-op in production. Would it be an issue for you in dev?

I feel like we need something to indicate that the `pathname` is expected to be fully encoded. The other suggestion, to throw, is interesting, but I thought might be a bit extreme.

pshrmn: (Also, please continue to make changes yourself where you see fit. My intention here wasn't to completely hijack your PR.)

OliverJAsh: Ah, if it's dev only, that would be fine.

If the `pathname` isn't encoded, what sort of problems will be occur? Sometimes it's good to fail early/loudly, but not sure if this is one of those or not.

pshrmn: The primary issue that I see is with path matching hard-coded encoded characters. If an app expects `pathname`s to be encoded, its paths should be encoded too:

```jsx
<Route path="/Beyonc%C3%A9" component={Bey} />
```

A location that comes from the browser will have an encoded `pathname` and everything willb e fine. However, if the user pushes a non-encoded location (e.g. `/Beyoncé`), the route would fail to match.

Being confident that the `pathname` always has the same form was a big part of why the decoding was originally added (alongside decoded params).

With throwing, I worry that someone would push a non-encoded location, the app doesn't catch it (or, worse, swallows it), and the user is left in limbo.

```jsx
<Link to="/Beyoncé">Beyoncé</Link>
// *click* and nothing happens
```

If we warn, but let the navigation go through, encoded route `path`s would fail (comparing `/Beyoncé` to `/Beyonc%C3%A9`), but paths with params would still work (e.g. `/artist/:name` would work for either `/artist/Beyoncé` or `/artist/Beyonc%C3%A9`). That would enable partial functionality, which may be worse than no functionality if it leads to subtle bugs.

```jsx
<Route path="/artist/:name" />

// location.pathname = /artist/Beyoncé
match.params.name === "Beyoncé

// location.pathname = /artist/Beyonc%C3%A9
match.params.name === "Beyonc%C3%A9"
```

I'm not really sure what the best approach would be, so I'm kind of hoping someone shows up and says "x is better than y for reason z".

pshrmn: Having typed that, I think that throwing would be more sound because it would not introduce bugs between encoded/non-encoded pathnames, but I would still like to hear input from others.

mjackson:
The last time I merged a PR that had to do with changing the way we do encoding, we ended up in this mess. @OliverJAsh Can you please summarize for me what your plan is here? There are so many changed files and so many comments...

pshrmn:
Just leaving a note that I reverted my changes for the time being because I don't want to hold this PR up while deciding what to do with non-fully-encoded pathnames.

modules/LocationUtils.js
OliverJAsh: Hmm, I think I agree.

> With throwing, I worry that someone would push a non-encoded location, the app doesn't catch it (or, worse, swallows it), and the user is left in limbo.

What if we didn't throw or warn? I guess this path just wouldn't match the route? In which case the user would presumably see a 404 (providing one had been setup).

On a side note, in your `params` example, I believe the actual behaviour should be:

```
<Route path="/artist/:name" />

// location.pathname = /artist/Beyonc%C3%A9
match.params.name === "Beyoncé"
```

`params` should be decoded (via `decodeURIComponent`), like they are in Express:

> decoding params using decodeURIComponent instead of the current decodeURI. This brings the behaviour inline with other frameworks such as Express, which automatically decodes the values in req.params using decodeURIComponent.

– https://github.com/ReactTraining/history/issues/505#issuecomment-375646353

OliverJAsh:
@mjackson 

> @OliverJAsh Can you please summarize for me what your plan is here?

The best summary is still the one I linked to in this PR's description: https://github.com/ReactTraining/history/issues/505#issuecomment-453175833. Essentially, with this change, we stop decoding `pathname` and just leave it as is. This fixes https://github.com/ReactTraining/history/issues/505.

There was [some discussion about whether to proactively encode](https://github.com/ReactTraining/history/issues/505#issuecomment-476222262), but it seems this was dropped due to issues.

There is still some ongoing discussion in this PR about whether to warn or throw when users provide `pathname`s which are not correctly encoded. However I'm not sure whether this should block this PR or not.

It's still not clear how to manage this upgrade with regards to React Router, but @pshrmn has [some thoughts](https://github.com/ReactTraining/history/pull/656#issuecomment-476201130.). @mjackson We will likely need your help on this.

When we do upgrade React Router to use this change, we will have to consider what to do with `match.params`. IIUC, they are currently decoded because they are taken from `pathname`, which was decoded, but won't be after this change. Thus we will have to decide whether or not React Router should decode `match.params`.

My two cents: it should decode them (using `decodeURIComponent`, _not_ `decodeURI`), because that's [what other frameworks like Express do](https://github.com/ReactTraining/history/issues/505#issuecomment-375646353). This will in turn fix another bug whereby not all characters are decoded inside `match.params`:
- https://github.com/ReactTraining/history/issues/505#issuecomment-375646353
- https://github.com/ReactTraining/react-router/issues/5607

modules/LocationUtils.js
pshrmn: In a future React Router release it should decode, but right now there is nothing to support this. We can't safely call `decodeURIComponent` on the matched params until we are certain that the `pathname` is fully encoded, so the change will have to come in a new RR version. Maybe we could fiddle with `<Route>` props to support this in a minor (`<Route decode />`), but we'll cross that bridge when we get there.

Michael and I are going to chat about this in a few hours. This change _is_ going to be made, the big question for me is determining what to do when we don't get a fully encoded pathname.

mjackson:
Thanks, @OliverJAsh. And thank you for sticking with this for so long. Let's get this merged.

Going forward, I think this library should have no opinion about encodings (it didn't, originally). When you read `location.pathname` it should be just like reading `window.location.pathname`. When you call `history.push(url)` it should work exactly like `window.history.pushState(undefined, undefined, url)` does.

Since this is a breaking change we will need to bump the major version. I'll add this to the roadmap.

OliverJAsh:
Thank you for merging this! Please ping me in any upstream React Router changes which relate to this? :-)

Dalzhim:
Great job @OliverJAsh on leading this PR! Thanks to @pshrmn and @mjackson for merging it! I'm looking forward for the upcoming release!

PR #659, opened 2019-01-26
Add basename as key on history object

jeffcarbs:
My team has built a few different React apps that live at different paths. For example:
- `example.com/users` is it's own React app
- `example.com/admin` is it's own React app

For each, the `basename` is defined at root component and I'm trying to write a shared helper to handle users "cmd + click" or "ctrl + click" that opens the URL in a new window or tab. For `Link` this is actually handled by default but my use-case is a user clicking on a `tr` element. That function looks like:
```js
const navigateOnClick = (href, history, window) => e => {
  if (e.altKey || e.ctrlKey || e.metaKey || e.shiftKey) {
    window.open(href);
  } else {
    history.push(href);
  }
};
```

The issue is that `href` will not include the basename. For the `else` case here (`history.push(href)`) this works as expected. However, for the `if` case (`window.open(href)`) it will go to the wrong URL (missing the basename). This PR just exposes `basename` on the `history` object which would allow me to do:
```js
const navigateOnClick = (href, history, window) => e => {
  if (e.altKey || e.ctrlKey || e.metaKey || e.shiftKey) {
    window.open(`${history.basename}${href}`);
  } else {
    history.push(href);
  }
};
```
so I wouldn't need to hard-code my basename which lets me reuse this helper.

### DIFF
--- a/modules/__tests__/BrowserHistory-test.js
+++ b/modules/__tests__/BrowserHistory-test.js
@@ -228,5 +228,10 @@ describe('a browser history', () => {
       const history = createHistory({ basename: '/prefix' });
       expect(history.location.pathname).toEqual('/');
     });
+
+    it('is available as key on history', () => {
+      const history = createHistory({ basename: '/prefix' });
+      expect(history.basename).toEqual('/prefix');
+    });
   });
 });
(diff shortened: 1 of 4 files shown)


jeffcarbs:
In reading through the file a bit more I actually found `createHref` which solves my problem. Here's the working code:
```js
const navigateOnClick = (href, history, window) => e => {
  if (e.altKey || e.ctrlKey || e.metaKey || e.shiftKey) {
    window.open(history.createHref({ pathname: href }));
  } else {
    history.push(href);
  }
};
```
Sorry for the noise 😅

PR #673, opened 2019-02-22
 Updated README.md with a history.block issue on iOS

freeboarder42:
On iOS (12) history.block does not work correctly when not using a custom prompt. window.confirm does not display the dialog when navigating away from a page, but instead just returns false. I described this non-solvable issue in README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,6 +203,8 @@ unblock();
 
 **Note:** You'll need to provide a `getUserConfirmation` function to use this feature with `createMemoryHistory` (see below).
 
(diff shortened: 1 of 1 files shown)


mjackson:
Thanks for the PR, @freeboarder42, but I think the issue may be that the history library doesn't currently support blocking navigations between different page reloads.

Can you please provide some more information? Maybe a demo app that we can use to reproduce the behavior you're seeing?

chaance:

PR #675, opened 2019-03-12
Inform users about known issues

Dalzhim:
Core functionnality of this library is broken in cases where URLs contain the %25 escape sequence. It is a major issue for users who need to handle any valid URL and it must be disclosed to those considering adoption of this library.


Dalzhim:
This PR is meant to document known issues with `history` versions prior to solving issue #505

OliverJAsh:
This is a good change until we've fixed #505.

PR #685, opened 2019-03-22
Add transformPathname option

pshrmn:
A potential solution to the decoding issues. This adds a function (currently named `transformPathname`) that is used to transform the `pathname` string during a `createLocation` call.

```js
function capitalizePathname(pathname) {
  return pathname.toUpperCase();
}
const history = createBrowserHistory({ transformPathname: capitalizePathname });
history.push("/hello");
history.location.pathname === "/HELLO"
```
The default behavior for browser and hash histories would be to set the `pathname` as the `href` of an anchor element and reading the anchor element's `pathname` property. This will give us the encoded version of the pathname. If the pathname is already encoded, it will do nothing.
```js
history.push("/test%20ing");
history.location.pathname === "/test%20ing"

history.push("/oh hello");
history.location.pathname === "/oh%20hello"

history.push("/Beyoncé");
history.location.pathname === "/Beyonc%C3%A9"
```
The memory history doesn't have to worry about encoding, so its default behavior is to return the provided string.

### Effects on React Router

This would be a breaking change for applications that use unicode characters in their paths. A "quick" fix would be for these applications to use a custom `transformPathname` function to make sure that `location.pathname` is a decoded string.

```js
// I haven't actually tested this!
const history = createBrowserHistory({
  transformPathname: pathname => decodeURI(pathname)
});
```

A more long term fix would probably be to encourage these applications to encode their paths.


pshrmn:
This wouldn't play nicely with the current `createLocation` export since creating a location would vary based on the history's `transformPathname` function. A `history` instance should be able to expose a `createLocation` function with the `transformPathname` function built-in.

Looking at the React Router source, `createLocation` is almost always used when we have access to the application's `history` instance. The only exception to this is the `StaticRouter`, but I think that that could be worked around.

pshrmn:
Maybe I'm overthinking this and _every_ history should use encoded pathnames internally.

I like using `transformPathname` more because it provides more freedom to let the user choose how their pathnames are formatted, but I'm not necessarily against using encoded pathnames everywhere.

pshrmn:
Putting this on hold, but may potentially re-open later.

PR #693, opened 2019-03-30
Fix/all paths key redundant

helios741:


helios741:
https://github.com/ReactTraining/history/pull/654

Issue #698, opened 2019-04-17
Ability to merge state changes instead of fully replacing


WesCossick:
We're working on the ability to easily store and recover state as a user navigates through a SPA. We started by using this package's `.replace()` function, which, as long as you're replacing the entire state, works fine.

But assuming there are multiple pieces of data that we want to store in the history state for a single location, and assuming the logic for storing each piece of data is isolated from one another, you need to merge the current state with what you want to change, and then pass the merged object to `.replace()`. That way you don't eliminate the other pieces of data.

However, because `.replace()` executes asynchronously, the state isn't fully updated right away. So subsequent calls to `.replace()` that try to merge using the current state end up using stale data (if those calls happen quickly enough).

Here's an example. Let's say you want to keep the following two pieces of data in the history state for a particular page:

```json
{
    "piece1": "original_value1",
    "piece2": "original_value2"
}
```

And let's say you want to update the state in two separate locations in rapid succession. I'll use [deepmerge](https://www.npmjs.com/package/deepmerge) for this example.

```js
// In one place in the code...
history.replace(location.pathname, deepmerge(location.state, {
    piece1: 'new_value1'
});

// In another place in the code...
history.replace(location.pathname, deepmerge(location.state, {
    piece2: 'new_value2'
});
```

At the end of this code, the state would contain:

```json
{
    "piece1": "original_value1",
    "piece2": "new_value2"
}
```

We're currently solving this issue by interfacing with the browser's history API directly. It appears to execute synchronously, so there's no timing issue. But since we're using React Router, it would be less hacky to interact directly with this package's API.

The easiest way to solve this, I believe, would be to introduce an option that asked for the state to be merged instead of fully replaced. If the merging happened at the right point in this package's codebase, it should eliminate the problem entirely. You could use [deepmerge](https://www.npmjs.com/package/deepmerge), even. It has no dependencies, so it wouldn't bloat the codebase.

StringEpsilon:
I'm not entirely convinced of the usecase, but I have a more versatile approach to solving it:

```replace``` could take a callback as an additional argument that is responsible of assigning the new state value based on the old and the new state.

```js
history.replace("/foo", { a: "foo" }); 
// location state = { a:  "foo" }
history.replace("/foo", { b: "bar" }, (old, new) => ({ ...old, ...new }) );
// location state = { a:  "foo", b: "bar" } 
``` 

That way, people can make write their own logic and history doesn't have a ~10% bundle size increase:

|    | min | minzip 
[...]

PR #699, opened 2019-04-20
Added callback to history.replace().

StringEpsilon:
Implements the feature request made in #698, specifically the idea I explained in my comment there.

I'm not entirely sure if the callback should also be added to other actions. A maintainers input either here or on the original feature request is greatly appreciated.

(Also: I can undo the white-space changes, as they create quite a lot of noise. But it's actually more consistent now)

### DIFF
--- a/modules/__tests__/BrowserHistory-test.js
+++ b/modules/__tests__/BrowserHistory-test.js
@@ -87,6 +87,12 @@ describe('a browser history', () => {
       });
     });
 
+    describe('replace state with callback', () => {
+      it('calls change listeners with the new location', done => {
+        TestSequences.ReplaceCallback(history, done);
+      });
+    });
+
     describe('location created by encoded and unencoded pathname', () => {
       it('produces the same location.pathname', done => {
         TestSequences.LocationPathnameAlwaysSame(history, done);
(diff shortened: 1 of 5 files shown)


mjackson:
I'm hesitant to add more API to this package, because it's already far too big. I think the root issue in #698 is that the location change happens asynchronously instead of synchronously like the browser does it. I think if we fix that (probably in a major bump) we can avoid adding more API to deal with the async behavior.

StringEpsilon:
@mjackson: is there active work on a 5.0 version? The roadmap issue is locked, so I can't comment on there.

If 5.0 is on the horizon, that's great news (and I'll close this PR). But if it's still a while off then this feature might help to bridge the gap for those in need. We can still rip it out in 5.0 anyway, if we properly communicate that it's a temporary solution til synchronicity is restored.

mjackson:
Everything on master is going into 5.0. I'm planning on releasing a beta this week. Just need to button up a few details. I'll update the roadmap.

mjackson:
> Everything on master is going into 5.0. I'm planning on releasing a beta this week.

Scratch that. I'm working on 5.0 in the `dev` branch. Development on 4.x will continue on `master`.

PR #702, opened 2019-05-16
PathUtils: New parsePath implementation.

### LINKED ISSUE #701: Reset search with question mark value to the empty string
Hi, I would like to suggest a feature here. Currently I'm using _react-router-dom_ for one of my projects, which of course uses _history_ dependency under the hood. Both _push_ and _replace_ methods have two possible overloads and may be executed whether with _path_ as string or _location descriptor_ as object. Using push method with _path_ works as expected and suppresses difference between f.e. `/about` and `/about?`, whereas push called with _location descriptor_ differentiate search values such as `""`(empty string) and `?` which causes redundant update issues in environment where you f.e. execute data fetching on search change, so that `location.search` equals to `?` after such navigate action. Wouldn't it be applicable to reset next search from `?` to `""` directly in history library to prevent such a case?

For more detailed example you might have a look on this [example](https://codesandbox.io/embed/llqx0q8v29) with react-router-dom. Just click the navigations links and observe how fetch counter is increased.

StringEpsilon:
* Slightly smaller build size
* Slightly faster (http://jsben.ch/zRUTx ~ 10%)
* Fixes #701

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,8 +1,8 @@
 {
   "esm/history.js": {
-    "bundled": 28364,
-    "minified": 12515,
-    "gzipped": 3683,
+    "bundled": 27886,
+    "minified": 12326,
+    "gzipped": 3587,
     "treeshaked": {
       "rollup": {
         "code": 208,
@@ -14,13 +14,13 @@
     }
   },
   "umd/history.js": {
-    "bundled": 33174,
-    "minified": 12021,
-    "gzipped": 3992
+    "bundled": 32668,
+    "minified": 11832,
+    "gzipped": 3900
   },
   "umd/history.min.js": {
-    "bundled": 30690,
-    "minified": 10155,
-    "gzipped": 3616
+    "bundled": 30184,
+    "minified": 9966,
+    "gzipped": 3516
   }
 }
--- a/modules/PathUtils.js
+++ b/modules/PathUtils.js
@@ -22,26 +22,18 @@ export function stripTrailingSlash(path) {
 }
 
 export function parsePath(path) {
-  let pathname = path || '/';
-  let search = '';
-  let hash = '';
-
-  const hashIndex = pathname.indexOf('#');
-  if (hashIndex !== -1) {
-    hash = pathname.substr(hashIndex);
-    pathname = pathname.substr(0, hashIndex);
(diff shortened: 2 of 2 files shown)


lanzhiheng:
I don't think it is a good idea. If you use this Regexp it will cut the content of the query string if it contain '?' character.

For example:

The url 'hello.com?abc=cut???mmmmm#match' we expect to get the result

``` js
{
  pathname:  'hello.com',
  search: '?abc=cut???mmmmm',
  hash: '#match'
}
``` 
But in your version

we will get 

``` js
{
  pathname:  'hello.com',
  search: '?abc=cut',
  hash: ''
}
```

It may cause some problem and sometimes Regexp will be very hard to maintain so I think it may not a good idea to do that even though it can improve the performance.

StringEpsilon:
It shouldn't cut this apart: `abc=cut???mmmmm`. But it works as intended for the hash part.

For the path `/foo/bar/baz?abc=cut???mmmmm#match`, this is the correct output:

```
{
  pathname:  '/foo/bar/baz',
  search: '?abc=cut???mmmmm#match',
  hash: ''
}
```

Will take a detailed look at the current behavior later.

lanzhiheng:
@StringEpsilon Why you said that `#match` is not a legal hash part?  In the origin code

```
const hashIndex = pathname.indexOf('#');
  if (hashIndex !== -1) {
    hash = pathname.substr(hashIndex);
    pathname = pathname.substr(0, hashIndex);
```

If will split the url string from tail to catch the hash value, and then catch the query sting. But in your new code you expect the url string like this.

``` js
> "http://hello.com#hash?q=querystring".match(/(\/?[^#?]*)(\#[^?]+)?(\?.+)?/)
[ 'http://hello.com#hash?q=querystring',
  'http://hello.com',
  '#hash',
  '?q=querystring',
  index: 0,
  input: 'http://hello.com#hash?q=querystring' ]
>
```

but I think the normal url with hash and query string should be `http://hello.com?q=querystring#hash`

StringEpsilon:
Oops. I misremembered the order of query and hash. My bad.

mjackson:
Thank you for the contribution, @StringEpsilon, but we aren't looking for these kinds of contributions right now.

PR #711, opened 2019-06-18
improve: warn when state is provided in HashHistory

laysent:
It's by design not allowed currently to provide `state` when push or replace in HashHistory. However, if `state` is provided in one argument (as property), there is no warning and state can be set.
Recommend to provide same behavior through all cases.


chaance:

mjackson:
Hash history is allowed to have state in v5. v5 uses pushState and replaceState instead of assigning to `window.location.hash` to change the location, so state is supported.

PR #713, opened 2019-06-29
Fixed a bug parsing URLs with both hash and search

natkuhn:
Since # always comes before ? in a URL, the code starting "const searchIndex" should come before "const hashIndex".

Currently the URL "/#foo?bar" is parsed to { pathname: "/", hash: "#foo?bar", search: "" }, not the desired { pathname: "/", hash: "#foo", search: "?bar" }.

I am such a newbie that I don't really know how to test my patch, but I'm pretty confident that it should work.

Here is a codesandbox illustrating the problem: https://codesandbox.io/embed/failing-history-rnqzu
Add "#foo?bar" in the demo browser address bar and hit enter. Then look at the console output.

Thanks!

### DIFF
--- a/modules/PathUtils.js
+++ b/modules/PathUtils.js
@@ -26,16 +26,16 @@ export function parsePath(path) {
   let search = '';
   let hash = '';
 
-  const hashIndex = pathname.indexOf('#');
-  if (hashIndex !== -1) {
-    hash = pathname.substr(hashIndex);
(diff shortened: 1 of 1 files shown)


natkuhn:
My mistake is explained in the issue, thank you. https://github.com/ReactTraining/history/issues/714

natkuhn:
My error is explained here: https://github.com/ReactTraining/history/issues/714
Thanks.

PR #727, opened 2019-08-12
Remove trailing slashes when trailingSlash is false

svedova:
This PR introduces a new property for `createBrowserHistory`. The reason of the PR is as follows:

1. If you have a basename like `/de`, or `/fr`
2. And your homepage is configured to display a language by default, even when the language is not present in the URL
3. And you're trying to link to your homepage

There will be a trailing slashed added to the url. The reason is simply because `createPath` will always return a `/` when the pathname is empty and `createHref` is simply appending the created path to the basename. 

This PR introduces the possibility to remove the ending slashes with a `trailingSlash` option passed to the `createBrowserHistory` function. By default it is true, which keeps the actual behaviour. When it is `false`, the `createHref` function will remove the trailing slashes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,7 @@ createBrowserHistory({
   basename: '', // The base URL of the app (see below)
   forceRefresh: false, // Set true to force full page refreshes
   keyLength: 6, // The length of location.key
+  trailingSlash: true, // If false, it will remove trailing slashes.
   // A function to use to confirm navigation with the user (see below)
   getUserConfirmation: (message, callback) => callback(window.confirm(message))
 });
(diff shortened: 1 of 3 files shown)


mjackson:
Can you please provide some code that shows what you are trying to describe in words? I really don't want to add an extra option to `createBrowserHistory` and an extra argument to `createHref` just for trailing slashes.

rporrasluc:
I have found this pull request when facing the same issue. Is there anyway to navigate to /en-gb for example when /en-gb is the basename? I have tried history.push/replace sending both / and "" and in both cases a trailing slash is added.

chaance:

PR #747, opened 2019-10-10
createBrowserHistory - fix revertPop

josheplibra:
createBrowserHistory - fix revertPop doesn't revert at the first time after reload page
// At the first time after page reload, toIndex = 0 and fromIndex = -1
// If set toIndex = 0, result of delta is toIndex - fromIndex = 0
// and then revert not work
// that mean address bar doesn't change back to "toLocation"
// so we ignore this case

### DIFF
--- a/modules/createBrowserHistory.js
+++ b/modules/createBrowserHistory.js
@@ -126,13 +126,19 @@ function createBrowserHistory(props = {}) {
     // keeping a list of keys we've seen in sessionStorage.
     // Instead, we just default to 0 for keys we don't know.
 
-    let toIndex = allKeys.indexOf(toLocation.key);
+    var toIndex = allKeys.indexOf(toLocation.key);
 
-    if (toIndex === -1) toIndex = 0;
-
-    let fromIndex = allKeys.indexOf(fromLocation.key);
-
-    if (fromIndex === -1) fromIndex = 0;
+    var fromIndex = allKeys.indexOf(fromLocation.key);
+    
+    // At the first time after page reload, toIndex = 0 and fromIndex = -1
+    // If set toIndex = 0, result of delta is toIndex - fromIndex = 0
+    // and then revert not work
+    // that mean address bar doesn't change back to "toLocation"
+    // so we ignore this case
+    if (!(toIndex === 0 && fromIndex === -1)) {
+      if (toIndex === -1) toIndex = 0;
+      if (fromIndex === -1) fromIndex = 0;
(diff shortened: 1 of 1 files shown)


sowka1995:
How can I use this fix in my project? I think that this trick will resolve my issue.

comp615:
One other fix that works is to generate an initial key:
https://github.com/ReactTraining/history/blob/v4/modules/createBrowserHistory.js#L71 becomes:
```
return createLocation(path, state, key || createKey());
```

I see a test that specifically tests against this:
https://github.com/ReactTraining/history/blob/v4/modules/__tests__/BrowserHistory-test.js#L45-L49.
But at the same time, it looks like v5 does set an initial key so maybe that's valid to update, I just don't know what the side-effects would be.

@mjackson would it be possible to pick one of the solutions given in this PR and put out a new v4 version? It's causing some issues in our Single Page App where we are asking users if they want to save their changes as a block, but then when they confirm, we are falling off this history stack because the block isn't properly re-popstating.

comp615:
Including Twitter repo here:
- Simulate Android in Chrome (this is due to us doing special back handling on mobile to let back close menus)
- Open https://twitter.com/charliecroom
- Click on an image in my pinned tweet
- Click the top right overflow menu

**Expected:**
- Hit browser back
- Menu closes and you are on the path for the image gallery
- Hitting back again goes back to the profile page

**Actual:**
- Hit browser back
- Menu closes and you are on the path for the profile page in your browser, but we're showing the image gallery
- Hitting back again falls off the nav stack and takes you to the new tab page.


**Cause:**
The fault on this specific bug is actually probably ours, the reason this happens is that we have some code on top of history to do things like go back as many times as needed to close a modal, and be aware of when we might fall off the stack (canGoBack). And to do that we reference location keys. If a location does not have a key, we assign it one (so it's easy to reason about as "undefined" as the first item is shady for truthiness tests). However, this causes history to internally drop the initial location off the stack since it no longer matches.

In short, if the initial item could have a location key as it does in v5, that helps a lot with us downstream being able to more definitely reason about things.

chaance:

chaance:
After having some time to think about this more and discuss it with @mjackson, I think we should probably add some documentation for cases like this rather than introduce a breaking change in v4 that we've already accounted for in v5.

That said, I'd love it if we had a place to capture these sort of nuances better. Might you be willing to contribute if I enable a wiki for this repo?

PR #748, opened 2019-10-17
ADD: install method

AlphaWong:
I find that I need to go to npm page (https://www.npmjs.com/package/history) to find the install method. It is better to have the install method


mjackson:
We already have installation instructions [in our docs](https://github.com/ReactTraining/history/blob/master/docs/Installation.md).

Issue #689, opened 2019-03-26
5.x Roadmap


mjackson:
This issue is a high-level roadmap for all work on the 5.x branch. The goal is to provide:

- a transparent process for the priority of a given piece of work
- an expected release version

The goal is not to discuss specific release dates, though some of that may happen.

This list is by no means exhaustive and will change from time to time as new issues are opened and PRs are merged.

### 5.0

- [ ] Stop decoding pathnames (see #656)

### Backlog

TODO

pshrmn:
Should blocking be removed in v5 or v6? If v5 removes blocking, then it wouldn't be compatible with RRv5 because of the `<Prompt>` component.

mjackson:
Excellent point, @pshrmn. We'll reserve the pared down version for v6. I have adjusted the version numbers to reflect this change.

pshrmn:
For v6, one possible change is to merge the `push` and `replace` functions into a single method. The default behavior would be to act like a link, so if the provided `to` is the same as the current location, it would replace and if they are not the same, it would push.

This would get rid of the issue with pushing the same location resulting in two session entries as well as simplify the API.

```js
history.navigate(to);
history.navigate(to, "PUSH");
history.navigate(to, "REPLACE");
```

I did this with my implementation and it looks like Ryan did a similar thing with reach. `state` is the big difference between the implementations; mine only allows `state` when `to` is an object, while with reach, `to` is always a string and the second argument is an options object for both state and the navigation type. I'm not sure what the best solution for `history` would be.

```js
// hickory
history.navigate({ pathname: "/somewhere", state: { success: true } });

// reach
history.navigate("/somewhere", { state: { success: true } });
```

danywheeler:
#520 #555 These resolves the same problems with different approach, the popstate is fired one page load on some old browsers like IOS safari below 9.3 and some other old webkit like.

#520 implementation requires a replaceState when page loads to distinguish if it's an page load.
#555 implementation is like delay list popstate the page load fired popstate always before document load, thus we delay popstate to skip the page load popstate event.

 @mjackson please consider implement this in v4.x or v5.0 use one of the provided solution I personally like the delay popstate  (#555 method) there shouldn't be any break change as this is the correct behavior(modern browser wouldn't fire popstate on page load!)

btimo:
Hey, been following the whole discussion around the decodeURI issue fixed by https://github.com/ReactTraining/history/pull/656. Would you have a timeline for the release of v5?

[...]

PR #750, opened 2019-10-28
Remove redundant '?' from search in createLocation

vimkin:
Here is a fix for an issue that is marked for 4.11 release in a following roadmap #689 and reported in the #701. It appears this issue is caused because of `createLocation` and not `createHref` as it is stated in the roadmap. The former was just copying it without modifying to the output. 

More in depth:
1. `Link` from `react-router-dom` calls `navigate` on click
```javascript
navigate() {
  // e.g. 'to' equals to { pathname: "/the/path", search: `?` }
  const location = resolveToLocation(to, context.location);
  const method = replace ? history.replace : history.push;

  // location descriptor is passed unchanged to corresponding history method
  // { pathname: "/the/path", search: `?` }
  method(location);
}
```
2. `history.push` or `history.replace` are called and here is some common code from theirs body
```javascript
// creates location from location descriptor passed from Link component
// path equals to { pathname: "/the/path", search: `?` }
const location = createLocation(path, state, createKey(), history.location);

// ... afterwards it sets location to the history state
setState({ action, location });
```
3. Since '?' wasn't cleaned anywhere it ended up in the next state and caused redundant refetches, rerenders etc.

### DIFF
--- a/modules/LocationUtils.js
+++ b/modules/LocationUtils.js
@@ -15,11 +15,14 @@ export function createLocation(path, state, key, currentLocation) {
 
     if (location.pathname === undefined) location.pathname = '';
 
-    if (location.search) {
-      if (location.search.charAt(0) !== '?')
-        location.search = '?' + location.search;
-    } else {
+    if (!location.search || location.search === '?') {
       location.search = '';
+    } else if (
+      typeof location.search === 'string' &&
+      location.search.charAt(0) !== '?'
+    ) {
+      // Prepends obligatory "?" to the search string if it's missed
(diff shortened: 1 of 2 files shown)


vadikace:
Super strange fail of the job with `TEST_ENV=source`: `Edge 17.17134.0 (Windows 10.0.0) a browser history push state calls change listeners with the new location FAILED`.  May it flaky CI bug because everything else is perfectly running without errors?

trymbill:
So ... this being merged anytime soon?

chaance:

Issue #745, opened 2019-10-07
decodeURI in code breaks my encoded URLs


vass-david:
Link to code that causes issues: https://github.com/ReactTraining/history/blob/7b8ab30f98855c31e74eb72f2592a8fff9d7d3ac/modules/LocationUtils.js#L36

I found that there were some issues raised a long while ago, but it seems that recently this has became an issue again.

My use case:
I'm receiving object names from API which are created by user. I can't control what's coming and I don't want to limit users from using special characters in their names. But even if I call `encodeURIComponent` on the name of an object, url will be decoded anyway and a user is navigated to wrong/faulty URL.

It's hard to tell what's the reason of `decodeURI` part of code, since commit name is "Undo changes from 78f016 and 0fe9cb" but github search wasn't able to find commits `78f016` or ` 0fe9cb`

Thanks for help

kaskelotti:
I'm working also with problems related to decodeURI. The commit for undoing changes stated in the message 

> This is a temporary change so we can get the master branch back into a
state to cut 4.x releases.

Now the question is, what is expected to happen and when? Is it best to fallback to using version 4.9.0 where this change was introduced?

@vass-david link to [78f016 commit](https://github.com/ReactTraining/history/commit/78f016f6e5b49e2a23ec41799faf160dd68d6b6b#diff-c0634631ce0ed08197ba5cf4643ca0cf)

michalkvasnicak:
We have this problem too, links where we need to have `100%` in them are completely broken.

dannsam:
I think it should be `decodeURIComponent` instead of `decodeURI`. Replacing it locally, fixes the issue for me. `decodeURI` is supposed to be used with full URIs https://stackoverflow.com/a/747700

kaskelotti:
This issue is pretty thoroughly discusses in #505 with possible solutions and workarounds, and what possible issues they might cause and why they might not be applicable in some specific cases. Please see that thread. We probably don't need another topic for that :) The discussion related to the problem has been ongoing for at least two years.

I think the real question is that why the fix to this issue was apparently taken out once it got introduced? Was there still something unexpected? And is it probable that there will be a fix-it-once-and-for all-solution or does every app just need their own work around to overcome this?

dannsam:
[...]

PR #751, opened 2019-11-05
v5

### LINKED ISSUE #624: createBrowserHistory with a specific window object
Inspired from this article https://hackernoon.com/using-a-react-16-portal-to-do-something-cool-2a2d627b0202 and trying to do a desktop Electron app which takes some seconds to initialize (parsing all own + vendor JS, fetching initial data from servers, etc.), I'm trying to spawn new windows and skipping this initialization, by using React portals.

For example, in a view I have a button "Contacts" which will popup the window with the Contacts component for the user to browse. By using this method, I can create an `about:blank` window, render everything in there by using a Portal, and that window will share the same state as the parent window.

In case that, for some reason, the popup window gets refreshed (the users presses Ctrl+R, or whatever), I can set the URL of the popup window by `replaceState` to a URL that I can define as a route, so when it gets refreshed it will go through all initialization (oh well...) and render the same component.

The problem I have though is if that Contacts component has some nested routes on it, as the code is running on the parent window AND the Router was initialized with it, all routes are based on the parent window, not on the popup window - So when I'm navigating the Contacts component, the URL gets replaced on the parent window which makes it harder to keep both window URLs in sync and it gets really impractical (because the parent window gets URL changes without changing any view)

If `createBrowserHistory` accepted a `Window` object as an option, then I could create a new history object every time I'm popping out a new window, and declare a new Router that uses that history object for that specific window.

### LINKED ISSUE #704: Add state to hash history
I'm currently in a similar position to that mentioned by @matthewrobb in #435, where I am limited to using the hash portion of the url for my dynamic application routing but I'm not limited in my support for older browsers and want to make use of storing state in the navigation whilst using hash urls.

I know #435 mentions a few workarounds to use `browserHistory` and `basename` but those workarounds feel like a bit of a hack and won't really work for me.

I am going to be putting together my own version of hashHistory that uses the browser history API and I was wondering if there would be any interest in adding that within to this library. I'd be happy to submit a PR to do it, but would need a bit of guidance on what approach to take here and what would need to be supported.

### LINKED ISSUE #723: supportsHistory is broken in Brave Browser
I'm using Brave Browser and hence this userAgent condition is failing

```
function supportsHistory() {
  var ua = window.navigator.userAgent;
  if ((ua.indexOf('Android 2.') !== -1 || ua.indexOf('Android 4.0') !== -1) && ua.indexOf('Mobile Safari') !== -1 && ua.indexOf('Chrome') === -1 && ua.indexOf('Windows Phone') === -1) return false;
  return window.history && 'pushState' in window.history;
}
```

My UA looks like this

`Mozilla/5.0 (Linux; U; Android 4.0; en-us; GT-I9300 Build/IMM76D) AppleWebKit/534.30 (KHTML, like Gecko) Version/4.0 Mobile Safari/534.30`

But it does support latest history apis.

Here's the original issue https://github.com/ReactTraining/react-router/issues/6844

### LINKED ISSUE #726: Is it still necessary reading hash from href now?
In the `createHashHistory.js`, there is a comment inner `getHashPath()`
```javascript
 // We can't use window.location.hash here because it's not
 // consistent across browsers - Firefox will pre-decode it!
```
But in the next function `pushHashPath`, it set hash directly.
```javascript
function pushHashPath(path) {
  window.location.hash = path;
}
```
I found the problem of pre-dcode location.hash happened about 6 years ago, and the recent version of firefox is working ok.

mjackson:
This PR is for version 5 which is a complete rewrite of version 4 that keeps most of its features while trimming a few that we don't really need. I'll keep this PR open and cut beta releases (npm tag `next`) until we're ready to release v5, at which point I'll merge to master.

New features:
- Remove legacy browser support (pre pushState)
- Add state to hash history
- Use custom `window` when creating history objects
- Better `history.block` API w/ `tx.retry` for retrying transitions
- Fix `location.pathname` encoding issues
- About 50% smaller
- No dependencies

Removed features:
- Removes basename support
- Removes relative pathname support in hash + memory histories
- Removes getUserConfirmation
- Removes keyLength
- Removes hashType

Fixes #624
Fixes #704
Fixes #723
Fixes #726


mjackson:
These test sequences [are failing in IE 11 and Edge](https://travis-ci.com/ReactTraining/history/jobs/253785931):

- [PushUnicodeLocation](https://github.com/ReactTraining/history/blob/dev/modules/__tests__/TestSequences/PushUnicodeLocation.js) – IE/Edge return unencoded pathname/search/hash
- [PushEncodedLocation](https://github.com/ReactTraining/history/blob/dev/modules/__tests__/TestSequences/PushEncodedLocation.js) – IE/Edge return unencoded pathname
- [LocationPathnameAlwaysSame](https://github.com/ReactTraining/history/blob/dev/modules/__tests__/TestSequences/LocationPathnameAlwaysSame.js) – IE/Edge return unencoded pathname when unencoded pathname is pushed

I'm thinking these tests are just bad tests and probably need to go. We should just provide the same pathname, search, and hash we get from the browser on our location objects and avoid `decodeURI` completely (same approach @OliverJAsh took in 930f0f7f). I believe this will fix the long-standing issue in #505 (and #745).

Any objections? /cc @timdorr @pshrmn

timdorr:
Do you mean v6? v5 is taken by 5.0.0

package.json
```diff
@@ -9,35 +9,21 @@
...
     "test": "karma start --single-run"
   },
   "dependencies": {
     "@babel/runtime": "^7.1.2",
-    "loose-envify": "^1.2.0",
-    "resolve-pathname": "^3.0.0",
-    "tiny-invariant": "^1.0.2",
-    "tiny-warning": "^1.0.0",
-    "value-equal": "^1.0.1"
+    "loose-envify": "^1.2.0"
```
timdorr: You can probably drop this. It's for Browserify users for the `process.env` stuff. Honestly, that doesn't belong in packages; it's a bundler problem. I'm not even sure why including it in the packages themselves fixes it. But if it needs to get fixed, that should happen on Browserify's end, not on literally _every_ package in npm that uses `process.env`.

timdorr:
Ugh, I am really not a fan of the flat source file. Flat bundles? Yes! Flat source? Not so much... 

It makes it harder to navigate through a nested set of functions when they're not broken out into individual files. With files, I can retain the scroll position of each one and just switch between tabs to navigate back and forth, or jump around. 

I do appreciate the drastically smaller code that you have to work with, so a single file becomes more practical (it would be untenable under v4), but I don't feel like it's reached that particular cutoff.

mjackson:
> Do you mean v6? v5 is taken by 5.0.0

RR is currently at v5, but history is still at v4. This next major will be history v5.

> I am really not a fan of the flat source file.

I've enjoyed working in a single file for this work, mostly because of all the code duplication between the implementations. Having them side by side makes it easy to compare them. But I'm not married to it. Also, I seem to remember webpack having a hard time doing tree-shaking when we had more than one module. Maybe this will make it easier for them.

Anyway, I was mostly asking for feedback about the encoding issues. I think those tests are bad and hoping we can just remove them and be done w/ encoding issues once and for all!

pshrmn:
I'll take a closer look this evening, but I agree about `history` not touching encoding.

timdorr:
Sorry, I deleted that comment about the version number because I thought this was the Router repo :P

Webpack only tree shakes along separate module files, IIRC. I've completely abandoned it for Rollup or Parcel now anyways, so that may be out of date info. Regardless, we're spitting out flat bundles anyways, so the size reduction stuff is a moot point. Just use a proper bundler like Rollup and don't worry about it. 

_Anyways_, I get what you're saying about code duplication reduction, but I would argue that's easier to see between multiple files. You can squint your eyes a bit, switch back and forth between some tabs, and literally see the abstraction visually. That's harder to do in a single file, IMHO. 

I will say this is nitpicky stuff (on my end), and I don't really have other comments about the more functional changes. It seems all pretty solid. And it's a good chance to knock out some long-standing bugs. :+1:

mjackson:
Just adding a note that I'd like to address #614 with the v5 release. Seems like setting `forceRefresh` at history object creation time isn't the best time to do it.

Andarist:
I'd like to raise a concern about module entry being minified - https://unpkg.com/history@5.0.0-beta.1/build/history.production.js 

It is a file that is going to be loaded **in development** and people (such as myself) might see this while debugging. I find it highly useful to be able to read the code while debugging (or even when just exploring), set up breakpoints and stuff.

Not only that but with current setup helpful `__DEV__` code gets stripped away, which should be done purely for production environments.

Happy to provide a PR fixing those, but we'd have to establish first how do we want to fix this.

mjackson:
The main idea of the "production" build is that all `__DEV__` code should be stripped. The idea is that you don't load that file in development, and you use the development build instead. It's [easy to do with CommonJS](https://github.com/ReactTraining/history/blob/39c4af9b7a07bbeaddd550a069abf3074373431b/index.js#L3-L6), but I don't know how to do this yet with ES modules.

Related: I noticed last night that closure compiler's ADVANCED_OPTIMIZATIONS mode is actually munging some of our public API, so I'm going to revert to using SIMPLE_OPTIMIZATIONS for now until I can figure out how to do the externs stuff properly.

Andarist:
I think this problem ultimately should be solved at the bundler level but there doesn't seem to be much interest in standardizing this right now 😢 

Due to the static nature of ESM the only thing that can be done is producing a similar "proxy" file that would do smth like this:
```
import * as prod from './prod'
import * as dev from './dev'

export const foo = process.env.NODE_ENV !== 'production' ? dev.foo : prod.foo
export const bar = process.env.NODE_ENV !== 'production' ? dev.bar : prod.bar
```
This approach has some challenges associated with it:
- you have to actually analyze the export shape to handle default export appropriately, that proxy file cannot try to reexport non-existent default export blindly
- you have to make sure your library tree-shakes nicely to avoid including some parts twice (as the content is evaluated eagerly for both dev & prod - which is different from CJS)

The former problem is not that difficult, just something to consider when implementing this approach. I think there might be some scripts that could be reused for this - I think I've seen smth like this in bundlephobia's codebase. 

The latter problem maybe could get "simplified" by additionally wrapping whole content of dev & prod in the appropriate NODE_ENV checks. This would look bizarre but would ensure that no duplicate code can get included and potentially would also result in initialization speed up for larger libraries. Actually something similar is already done by React ( https://unpkg.com/react@16.11.0/cjs/react.development.js - notice how everything inside dev bundle is wrapped by NODE_ENV check, just to be sure that no development code can get shipped in production). To create a reusable script for generating files like this we'd have to consider few edge cases/more complex scenarios - but my hot take is that this is very much possible to create.

mjackson:
I agree this should be handled by the bundlers. My thinking was that since none of them differentiate between dev and prod sources, we should default to prod and then provide instructions for people who want a better dev experience.

For example we could say "if you want a better dev experience with webpack, use an alias in your `webpack.config.js` like this":

```js
const historyAlias = require('history/webpackAlias');

module.exports = {
  "resolve": {
    "alias": {
      "history": historyAlias
    }
  }
}
```

Main drawback here is that it requires a bit of work to get a nicer dev experience, but I'm willing to make that trade-off.

Andarist:
That ofc can be done, but the discoverability would be low and wouldn't cover 0-config tools like CRA. WDYT about my approach with excessive NODE_ENV checks? I understand that this sounds like something completely out of the scope of this repository, but I could create a package (rollup plugin maybe?) that would do this. Would you be willing to use it in such a scenario?

mjackson:
Yes, I think we'd be waking to take that approach. 

Another thing I'm wondering is why we don't just leave the `process.env.NODE_ENV` checks in the code and skip minification entirely. The code wouldn't work w/out a bundler, but anybody who is using either `main` or `module` is already using a bundler. We could also provide standalone dev/prod builds for people who want them, but they would not be referenced in main/module. Wdyt?

Andarist:
> Another thing I'm wondering is why we don't just leave the process.env.NODE_ENV checks in the code and skip minification entirely.

That is what I'm generally doing in other projects. The only possible advantage that I can see of minification being "included" in distributed files (which are meant to be consumed by bundlers anyway) is that in theory we, as maintainers, can setup minification more effectively than our consumers. From my experience though this rarely brings any real gains though.

mjackson:
Everybody is already running their production code through minifiers anyway, and setting `process.env.NODE_ENV=production` is a well-known technique that is built into webpack's production mode. So I'd vote we just skip minification on our side. I'd propose we ship the following builds:

- `history.js` - the `main` entry in package.json, CommonJS exports, includes `process.env.NODE_ENV` checks
- `history.mjs` - the `module` entry in package.json, ESM exports, includes `process.env.NODE_ENV` checks

We *could* also ship a few pre-built files that don't have the `process.env.NODE_ENV` checks (and are maybe even minified) but I'd expect these would be used a lot less. Examples:

- `history.development.js` - CommonJS exports, includes all __DEV__ code
- `history.production.min.js` - CommonJS exports, no __DEV__ code, minified
- `history.development.mjs` - ESM exports, includes all __DEV__ code
- `history.production.min.mjs` - ESM exports, no __DEV__ code, minified

I'd probably put `history.production.js` in our `unpkg` field in package.json.

WDYT @Andarist?

Andarist:
I believe this is a good direction - optionally for CJS you could have a "proxy" file in `history.js` like this one:
https://unpkg.com/react@16.12.0/index.js
As for CJS it's super easy to create such.

I wouldn't care about some of those files, but there is also not much harm in providing them (for example `history.development.mjs` - hard to imagine anyone using it as it would have to be manually chosen).

> I'd probably put history.production.min.js in our unpkg field in package.json.

This also makes sense for me 👍

timdorr:
I thought we already did have the proxy file? https://unpkg.com/browse/history@4.10.1/index.js

mjackson:
@timdorr Yep, we already have that file 👍 

In the end I went with individual builds for each history type + dev/prod UMD builds for anyone still using CommonJS or a script tag. So we have the following files:

- `browser.js` - browser history only, ESM, includes `process.env.NODE_ENV` checks
- `hash.js` - hash history only, ESM, includes `process.env.NODE_ENV` checks
- `memory.js` - memory history only, ESM, includes `process.env.NODE_ENV` checks
- `history.js` - all 3 (not recommended), ESM, includes `process.env.NODE_ENV` checks
- `history.development.js` - all 3, UMD, includes all `__DEV__` code
- `history.production.min.js` - all 3, UMD, no `__DEV__` code, minified

I think that covers all our bases. Most people using bundlers will just `import history from "history/browser"` and be good to go.

mjackson:
OK, I just pushed another release that includes some "web modules" builds, which are meant for being used in a `<script type=module>`. I also updated [the Installation docs](https://github.com/ReactTraining/history/blob/dev/docs/Installation.md), so please let me know if you read that and something doesn't make sense.

package.json
```diff
@@ -1,56 +1,30 @@
...
+  "name": "history-packages",
+  "private": true,
   "scripts": {
-    "build": "rollup -c",
+    "build": "node ./scripts/build.js",
+    "size": "filesize",
     "clean": "git clean -fdX .",
-    "lint": "eslint modules",
+    "lint": "eslint .",
     "prepublishOnly": "yarn build",
```
Andarist: is this prepublishOnly needed here? seems not necessary for a private monorepo root

Andarist:
Some feedback:
- the structure of the repository is currently a little bit confusing for me. It's a "monorepo" with a single package and its dist files are put into a root build directory. I had to put some thought into understanding this and tracing how things tie together. Most commonly dist files are put close to source files - it's IMHO easier then to inspect the package as a whole because one only has to focus on a single directory.
- I don't quite think my concern about dist files readability & debuggability was addressed. Maybe it's just me - but I have a terrible experience with source maps and prefer debugging transpiled code (what actually get executed by a browser and not what I think it executes - the source code)
- I believe that `createBrowserHistory` imported from `history` package should be the same as the one imported from `history/browser`. Right now if one imports both they will have duplicated code in their bundles.
- There are no CJS entries for sub-entries like `history/memory`. From what I believe this can be used in node. In general, I would advocate for having the very same exports shape and formats for all entries. It might be confusing that some are ESM-exclusive but some are operating in a "dual" mode.

salvoravida:
Hi, just a question about :
* Removes basename support

What happens to React Apps that have a basepath?

packages/history/modules/index.js
```diff
@@ -0,0 +1,696 @@
...
+      } else {
+        applyTx(nextAction);
+      }
+    }
+  }
+
+  window.addEventListener(PopStateEventType, handlePop);
+
+  // TODO: Is this still necessary? Which browsers do
+  // not trigger popstate when the hash changes?
```
StringEpsilon: According to the MDN, popstate does not fire on hash changes for IE11 and old (trident) Edge versions.

If you want to support IE11, you have to keep the workaround.

https://developer.mozilla.org/de/docs/Web/API/Window/popstate_event

mjackson: That's what I thought. Thanks for confirming :)

packages/history/history.d.ts
```diff
@@ -0,0 +1,62 @@
...
+export function createHashHistory(options?: { window?: Window }): History;
+
+type InitialEntry = Path | PathPieces;
+export function createMemoryHistory(options?: {
+  initialEntries?: InitialEntry[];
+  initialIndex?: number;
+}): MemoryHistory;
+
+export function createPath(pieces: PathPieces): Path;
+export function parsePath(path: Path): PathPieces;
```
StringEpsilon: Given that parsePath gracefully handles an undefined or null `path` parameter, wouldn't it make more sense to mark path as optional?

```ts
export function parsePath(path?: Path | null): PathPieces;
```

As is, you kinda force to handle possible null / undefined values by the caller when they otherwise don't need to.

packages/history/history.ts
```diff
@@ -0,0 +1,983 @@
...
+   * The URL fragment identifier, beginning with a #.
+   */
+  hash?: string;
+};
+
+/**
+ * An object that is used to associate some arbitrary data with a
+ * location, but that does not appear in the URL path.
+ */
+export type State = object;
```
OliverJAsh, line 31: Hey, not sure if you saw this, but I left a comment in the original [TS PR](https://github.com/ReactTraining/history/pull/774) about whether this type is correct: https://github.com/ReactTraining/history/pull/774#discussion_r376315739

packages/history/history.ts
```diff
@@ -0,0 +1,983 @@
...
+  usr?: State;
+  key?: string;
+  idx: number;
+};
+
+const BeforeUnloadEventType = 'beforeunload';
+const HashChangeEventType = 'hashchange';
+const PopStateEventType = 'popstate';
+
+const readOnly: (obj: any) => any = __DEV__
```
OliverJAsh, line 259: Hey, not sure if you saw this, but I left a comment in the original [TS PR](https://github.com/ReactTraining/history/pull/774) about whether this type is correct: https://github.com/ReactTraining/history/pull/774#discussion_r376317031

PR #753, opened 2019-11-18
Use Closure Compiler with Advanced Mode

kristoferbaxter:
Also adds size tracking via `filesize`.


mjackson:
Awesome work, thanks @kristoferbaxter! 🙌

PR #761, opened 2019-12-19
Remove custom closure annotations

kristoferbaxter:
These options were increasing the build size – since the repo uses Babel for transpilation based on `preset-env` leaving the default options from the closure compiler plugin (language_out=NO_TRANSPILE, etc) will net a smaller output.


mjackson:
Superseded by 5f723464ecf2582d7f346c64a73ea90495d827e9

PR #769, opened 2020-01-22
Add Support for Custom Confirmation Dialogue Box

raynorelyp:
history.block can set the dialogue for confirmation window, but if you have multiple confirmation windows for different pages, you need to recreate the history object. This change allows you to pass a config object down to your custom getUserConfirmation function to help configure your confirmation window you show.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,8 +1,8 @@
 {
   "esm/history.js": {
-    "bundled": 28076,
-    "minified": 12353,
-    "gzipped": 3575,
+    "bundled": 28727,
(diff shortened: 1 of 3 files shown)


mjackson:
I'd rather not expand the v4 API to fix this particular problem. The v5 API is much more straightforward and allows you to setup blocking behavior at the `history.block` call site instead of upon creation of the `history` object.

v5 is currently in beta. Please see [the v5 docs on blocking](https://github.com/ReactTraining/history/blob/dev/docs/blocking-transitions.md) on the dev branch for more details.

Issue #814, opened 2020-07-08
`push()` and `replace()` will keep current `search` and `hash` string


yangmingshan:
**When**
Current location is `http://localhost?a=a#b`

**Do**
```js
history.push('/');
// OR
history.replace('/');
```

**Expect**
Location change to `http://localhost`

**Actual**
Location still is `http://localhost?a=a#b`

arnaudhuc:
Got the same issue

jnields:
Yeah, this definitely seems like a bug to me too as it's a breaking change that is not documented with the 5.0.0 release, and just feels very unintuitive.

https://github.com/ReactTraining/history/releases/tag/v5.0.0

GordonSmith:
I think what I am seeing the same issue, but with "createHashHistory":

1. My URL is:  `http://localhost:8080/esp/files/index.html#/main?lastNDays=31&protected=false&state=completed`
2. Use edits the URL in the address bar to:  `http://localhost:8080/esp/files/index.html#/main?lastNDays=999&protected=true`
3. In `history.listen` I still get:  `?lastNDays=31&protected=false&state=completed` for the search...

Hypnosphi:
Here's my workaround:

```js
const push = (to, state) => history.push({
  hash: '',
  search: '',
  ...(typeof to === 'string' ? parsePath(to) : to)
}, state)
```

zhangdingding:
It looks more like a bug than a feature 🤔

unbugx:
same is here #808

PR #818, opened 2020-07-17
Clear previous search and hash on push/replace with pathname only

sachmata:
When `history.push` or `history.replace` is called with string URL without `search` or `hash` part the ones from previous location are kept.

```
history.push('https://reacttraining.com/with-query?query=hello');

// https://reacttraining.com/with-query?query=hello -> This is OK
console.log(history.createHref(history.location));

// new route is pushed without search
history.push('https://reacttraining.com/no-query');

// https://reacttraining.com/no-query?query=hello -> Current result keeps the old search
console.log(history.createHref(history.location));
```

This change adds default empty values for `search` and `hash` which are replaced if `parsePath` function finds some.

If the previous `search` or `hash` have to be reused `PartialPath` object can be passed with omitted keys.

### DIFF
--- a/packages/history/__tests__/TestSequences/PushNewLocation.js
+++ b/packages/history/__tests__/TestSequences/PushNewLocation.js
@@ -20,6 +20,18 @@ export default (history, done) => {
         state: null,
         key: expect.any(String)
       });
+
+      history.push('/other');
+    },
+    ({ action, location }) => {
+      expect(action).toBe('PUSH');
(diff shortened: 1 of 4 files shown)


jaydenseric:
Fixes https://github.com/ReactTraining/history/issues/814 ?

sachmata:
> Fixes #814 ?

Yes, I have missed that is already reported.

Hypnosphi:
@mjackson can you please review?

PR #820, opened 2020-07-18
createPath: Prefix search and hash parts as needed.

### LINKED ISSUE #813: createPath no longer prefixes "search" and "hash" parameters
During the transition to typescript, `createPath` function has changed from this:

```
function createPath(location) {
  var pathname = location.pathname,
      search = location.search,
      hash = location.hash;
  var path = pathname || '/';
  if (search && search !== '?') path += search.charAt(0) === '?' ? search : "?" + search;
  if (hash && hash !== '#') path += hash.charAt(0) === '#' ? hash : "#" + hash;
  return path;
}
```

to this

```
export function createPath({
  pathname = '/',
  search = '',
  hash = ''
}: PartialPath) {
  return pathname + search + hash;
}
```

You are no longer ensuring that passed `search` and `hash` are prefixed, which kind of broke our site.

You should either restore the old behavior or document the breaking change.

StringEpsilon:
Fixes #813 by restoring the version 4 behavior.

This is technically a breaking change, but my interpretation is that the breaking change from v4 to v5 was not made intentionally. So this is just restoring the old (desired) behavior.

### DIFF
--- a/packages/history/index.ts
+++ b/packages/history/index.ts
@@ -1080,6 +1080,12 @@ export function createPath({
   search = '',
   hash = ''
 }: PartialPath) {
+  if (search && search[0] !== '?'){
+    search = '?' + search;
+  }
+  if (hash && hash[0] !== '#'){
+    hash = '#' + hash;
+  }
   return pathname + search + hash;
 }
 


StringEpsilon:
Also fixes ReactTraining/react-router/issues/7496

Issue #638, opened 2018-11-02
Does force refresh also apply to user's back button?


reywright:
I'm having an issue in my application where I've configured forceRefresh, and this works perfect when `push` and other stuff is used, but if the user presses the back button after they've navigated, they navigate without full refresh. Any ideas?

history 4.7.2

ColeWoodford:
I am also wondering about this. Pressing the browser's back button does not seem to cause a full refresh.
Also using 4.7.2

pshrmn:
If you open your browser's dev tools and inspect network requests, there is no request to the server when you click the back button?

`history` does not explicitly handle `forceRefresh` in its `popstate` event listener, but it shouldn't have to. When you `push` with `forceRefresh`, the pushed page will have a different `document` from the previous page (`pushState` reuses the same document). Clicking the back button will have to unload the current `document` and load a `document` for the previous location, which should send a request to your server.

Out of curiosity, why do you need to use `history` if you are treating the application as a multi-page application? For example, if you are using React Router, you could achieve the same effect by using anchors instead of `<Link>`s. That would let the browser handle all navigation for you.

reywright:
We'd like the ability to turn it off dynamically, as the user experience between the two is a tad bit different based on the domain of application. It's much much easier to depend on force reload and do pushes on all links, rather than have to sometimes use push, sometimes use anchors.

pshrmn:
Makes sense, thanks! I just threw a repo together and see that it isn't reloading. I'll see if I have the time to take a little deeper dive and figure out why it isn't.

pshrmn:
It turns out that `history.push` calls [`window.history.pushState`](https://github.com/ReactTraining/history/blob/82686d45e6a369a01506276d0b8896d845430e48/modules/createBrowserHistory.js#L180) even when `forceRefresh` is `true`. When you call `location.href = href` (or `location.assign()`) with the same URL as the current URL after calling `pushState()`, the document must get shared despite the hard load.

If I move the `pushState` call linked above to only be called when not force refreshing, then popping also does a full reload.

If you would like to open a PR to make this change, that would be appreciated.

reywright:
@mjackson any reason you closed this issue?

mjackson:
Hey, sorry @reywright. I must've closed it accidentally. Re-opening.

reywright:
I'm going to go ahead and close this. Please file another issue if you want to take a shot at getting this fixed.

PR #842, opened 2020-11-10
Force refresh on browser back navigation

banada:
Calling globalHistory.pushState when forceRefresh is enabled
causes the document to be loaded from cache instead of reloaded.

This breaks forceRefresh when navigating backwards using the
browser's navigation buttons.

Now, we only call pushState if not in forceRefresh mode.

For more information, see this issue:
https://github.com/ReactTraining/history/issues/638#issuecomment-435545828


vikas-0:
Can this be merged?

facundocoto (approved):

vikas-0 (approved):

PR #869, opened 2021-04-29
Replace 'partial' ts interface implementation with built-in Partial utility type

http417:
This is very minor refactor that removes redundant partial interface implementation with built-in typescript utility Partial types.


mjackson:
Duplicate of #839

PR #892, opened 2021-08-17
Support numeric `to`

chaance:
Resolves the same issue as https://github.com/remix-run/history/pull/795, but quite a bit seems to have changed since that PR was opened. We need to tweak some of our type checking logic if we're going to parse numbers as potential paths.


chaance:
Thinking about this more, I don't think we should support this. Yes it's convenient but `String(to)` is not a big workaround, and if we say that `number` is an acceptable type we'd probably want to do `NaN` checks, and honestly I think that's complexity we can leave to the consumer if it's important to them.

PR #902, opened 2021-11-01
fix: search params persisting on redirects

jacob-ebey:


packages/history/index.ts
```diff
@@ -1051,7 +1051,9 @@ export function createPath({
  * @see https://github.com/ReactTraining/history/tree/master/docs/api-reference.md#parsepath
  */
 export function parsePath(path: string) {
-  let partialPath: PartialPath = {};
+  let partialPath: PartialPath = {
+    search: ''
+  };
 
   if (path) {
```
ryanflorence, line 1058: what the heck?

PR #908, opened 2021-11-18
fix: #907

### LINKED ISSUE #907: v5 createPath issue with null values. was fine in v4
push({pathname:'login', hash:null})  => url created "loginnull"

the problem is with the history v5:
https://github.com/remix-run/history/blob/main/packages/history/index.ts#L1052

it was fine with history v4
https://github.com/remix-run/history/blob/v4/modules/PathUtils.js#L56

related issue: https://github.com/salvoravida/redux-first-history/issues/53

liuhanqu:
Fix #907

### DIFF
--- a/packages/history/index.ts
+++ b/packages/history/index.ts
@@ -1049,6 +1049,9 @@ export function createPath({
   search = '',
   hash = ''
 }: PartialPath) {
+  if (hash && hash !== '#') {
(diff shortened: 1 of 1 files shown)


brophdawg11:
Closing as this should be fixed in `5.2.0` - please see https://github.com/remix-run/history/issues/907#issuecomment-1054605323

Issue #925, opened 2021-12-31
state: unknown Object is of type 'unknown'.


sathiya-pyramidions:
`type of state is unknown` 

![Screenshot (10)](https://user-images.githubusercontent.com/89627143/147808877-39065c2b-a123-42d4-863d-333104071617.png)

![Screenshot (9)](https://user-images.githubusercontent.com/89627143/147808869-74860e08-946e-4793-8e69-41451ea167d3.png)

PR #926, opened 2021-12-31
Fix for #925

sathiya-pyramidions:
This PR fixes the issue #925 where unknown is used as type instead of object | null / any.


timdorr:
Thanks, but this was an intentional change in 5.2.0: https://github.com/remix-run/history/releases/tag/v5.2.0

PR #931, opened 2022-01-18
fix: allow `import {xyz} from "history"`

jacob-ebey:
chore: simplify build config
chore: move type generation out of rollup
chore: use babel instead of another plugin for transpiling TS
chore: use typescript references for easier migraiton later
chore: bundle babel helpers
chore: remove dep on @babel/runtime


ryanflorence:
closing in favor of #934

PR #932, opened 2022-01-20
[wip] Changes for extracting new router package

jamiebuilds:
- Newer versions of Prettier have different defaults for `arrowParens` so I added `arrowParens: avoid` to your config so it wouldn't change the code style if someone is running a newer version of prettier
- Removed all of the `block()` related code
- Replaced usages of `substr()` (which is deprecated) with `substring()`, this didn't change how you were using it
- Removed all of the `action` related code. I can revert this if the scroll restoration work does end up needing it
- Removed all of the index tracking in a separate commit, it was only needed for `block()`


packages/history/index.ts
```diff
@@ -1008,40 +715,29 @@ function clamp(n: number, lowerBound: number, upperBound: number) {
...
-
 type Events<F> = {
-  length: number;
   push: (fn: F) => () => void;
   call: (arg: any) => void;
 };
 
 function createEvents<F extends Function>(): Events<F> {
-  let handlers: F[] = [];
+  let handlers = new Set<F>();
```
jamiebuilds, line 724: Oh, also I don't know what kind of browser support you want, but I replaced the `createEvents` implementation with one that uses a `Set` which has limited support in IE 11 (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set#browser_compatibility)

jamiebuilds:
# Size diff

```
history.production.min.js (gz) - Before: 2.65 kB - After: 1.19 kB
```

PR #934, opened 2022-02-04
fix: import history from ESM

ryanflorence:
Removing closure compiiler fixes the problem. The build is now much faster and the resulting file size difference is marginal.

The UMD production file size went from 6,105 bytes with closure compiler to 6,164 without it. We can afford 59 bytes to get ESM support.

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,10 +65,10 @@
   },
   "filesize": {
     "build/history/history.production.min.js": {
-      "none": "5 kB"
+      "none": "5.06 kB"
     },
     "build/history/umd/history.production.min.js": {
-      "none": "6 kB"
+      "none": "6.03 kB"
     }
   }
 }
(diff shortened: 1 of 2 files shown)


ryanflorence:
Tests run for me locally. I also verified importing works with this script:

```js
npm run build
mkdir ~/Desktop/test
touch ~/Desktop/test/go.mjs
cp -R build ~/Desktop/test/node_modules
```

Then running this script works as expected:

```js
import { createMemoryHistory } from 'history';
console.log(createMemoryHistory);
```

kentcdodds (approved):
LGTM

mcansh (approved):

mjackson (approved):

kristoferbaxter:
Seems like a smart tradeoff to me as well.

PR #944, opened 2022-02-28
feat: remove block() API

brophdawg11:
Re-implements selective portions of #932 on top of the latest `dev` (post-prettier update) (does not remove `action` logic).

* Removes the `block()` API
* Removes associated tests for the `block()` api
* Removes `index` logic from `createBrowserHistory`/`createHashHistory` as it was only used for `block()`

BREAKING CHANGE: The removal of `history.block()` will be a breaking change in `history@6`


mjackson:
Thanks for doing this exploratory work, @brophdawg11. I'm going to close this and we'll follow up in the router repo. 👍

sdlbp:
Why is this API being deleted? What factors were considered?

PR #945, opened 2022-02-28
chore: extract shared createUrlBasedHistory method

brophdawg11:
### Dependent on #944

Leverages work from #932 to abstract a shared `createUrlBasedHistory` for use by browser/hash versions


mjackson:
Again, let's follow up in the router repo 👍

Issue #921, opened 2021-12-22
Difficult to migrate .block() use case to v5 


liamqma:
Hi,

I use history.block to check whether the next location can be transition to. If not, do a full page load. The pseudo code looks like

```
const unblock = history.block((tx) => {
  // canTransition is the function that checks if next location can be SPA transition to
  // If false, do a full page load on next location's path
  if (!canTransition()) {
    return window.location.assign(nextPath);
  }
  unblock();
  tx.retry();
});
```
Full code example: https://codesandbox.io/s/distracted-river-8yczz?file=/src/index.js

So depending on the result of canTransition(), it should either unblock the blocker or do a full page reload. Users shouldn't notice much difference. This is fine except for the navigation that reloads the page. Users see something like this (Chrome):
![image](https://user-images.githubusercontent.com/4413219/147031018-002f170b-290b-4474-afca-4242cc6baf05.png)
It doesn't make sense to our users as there isn't any unsaved changes.
I've read the [documentation](https://github.com/remix-run/history/blob/main/docs/blocking-transitions.md) and understand the library registers a beforeunload handler to prevent the navigation. Maybe v5's .block is not designed for my use case.

But I am able to achieve this use case via history v4. I've also tried history.listen but .listen is too late as it happens at the end of `applyTx`.

Any suggestion is much appreciated. 🙂
The solution I can think of is to remove `window.addEventListener(BeforeUnloadEventType, promptBeforeUnload);` from the library and let consumer of the library to handle it.

liamqma:
I've found a hacky workaround 😬

```
window.addEventListener('beforeunload', () => {
        unBlock();
});
```

Register our own BeforeUnloadEvent and clean up our transition blockers before history runs its check

dubzzz:
@liamqma Thank you so much for the workaround it works perfectly well in my case too. I just had to register my own event listener before calling block which I did the other way initially.

Meanwhile I am wondering if calling the provided blockers on "beforeunload" could be something directly implemented inside history itself. I came up with this code on my side (highly inspired from react-router beta.6):
```ts
// location being undefined when beforeunload cases
export type BlockerFunction = (tx: { location: Location | undefined; retry: () => void }) => void;

export function useBlocker(blocker: BlockerFunction): void {
  const navigationContext = useContext(UNSAFE_NavigationContext);
  const navigator = navigationContext.navigator as History;

  useEffect(() => {
    const beforeUnloadCallback = () => {
      let closeAuthorized = false;
      blocker({
[...]

PR #949, opened 2022-04-20
Allow adding blockers that don't create a beforeUnload event

karlshea:
This allows adding blockers that don't create a beforeUnload event, so blocks can prevent or otherwise handle location changes but won't block reloading the page.

Normal blockers are looked at first (to support prompts), and then if they all allow the transaction the no-unload blockers are checked afterwards.

See #921 for reasoning (forcing a full page load for non-SPA routes). What are the maintainer's thoughts on this approach?

### DIFF
--- a/packages/history/__tests__/TestSequences/BlockEverythingNoUnload.js
+++ b/packages/history/__tests__/TestSequences/BlockEverythingNoUnload.js
@@ -1,0 +1,25 @@
+import expect from "expect";
+
+import { execSteps } from "./utils.js";
+
+export default (history, done) => {
+  let steps = [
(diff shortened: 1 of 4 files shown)


ofirgeller:
Could also add a flag to block. would like to see some solution merged, I had to drop the blocking feature on an app so i can upgrade to react-router v6 which sucks :(

tobi-fis:
I don't understand how this is not merged yet. The unavoidable beforeunload handler forces you to implement hacky workarounds if you simply want to intercept navigations by using `push`, without the misleading browser popup showing up.

PR #939, opened 2022-02-15
Restore entries attribute in MemoryHistory interface

Donorlin:
Would feature fix this https://github.com/remix-run/history/issues/828

This PR gives back the feature of `MemoryHistory.entries` which was available in version 4

Many people are dependent on this feature, mainly because it gives more control over memory history itself.

For example, we are running SPA in the SSR (company historical reasons). This means the SPA state is lost whenever a site is reloaded. In version `history@4` we are able to save entries from the history object and pass it to `initialEntries` of the `MemoryRouter` to seed it back to its previous state (controlled behavior).

In react-router@6 we are still able to access the `MemoryHistory` object with `UNSAFE_NavigationContext`, but not its entries. So we are not able to implement a controlled `MemoryRouter`.

If it was intentional to remove entries for API reasons, close this PR.

Thank you for your response.

PR #951, opened 2022-04-29
set initial memory history key to "default"

fmauquie:
Documentation states that the initial entry always has key `"default"`. However on memory history this is not the case.

We check that the key is `"default"` to set a default behavior when the user clicks a "back" button to go back to home, but arrived at the page from a direct link. This works perfectly in the browser, but failed in the test because the key was not "default" on MemoryRouter.


Nantris:
I hope this could get merged soon. There are various, now rather long-standing, but extremely simple issues with MemoryHistory which don't seem to be moving forward, which makes upgrading to React Router 6 impossible, which breaks React.StrictMode. It's a cascade of consequences from like eight lines of missing code between this issue and #939.

